PR #47, opened 2016-05-03
Use HTTP Error for HTTP Errors such as 4xx and 5xx

rhyshort:
## What 
Move away from using Error objects such as `.GetDocumentFailed` to using `.HTTP` error instead.

## Testing 
No new testing.

## Reviewers
reviewer @ricellis

### DIFF
--- a/Source/DeleteDatabaseOperation.swift
+++ b/Source/DeleteDatabaseOperation.swift
@@ -76,7 +76,7 @@ public class DeleteDatabaseOperation : CouchOperation {
                 response = nil
             }
             
-            self.deleteDatabaseCompletionHandler?(statusCode: statusCode, operationError:Errors.CreateDatabaseFailed(statusCode: statusCode, jsonResponse: response))
+            self.deleteDatabaseCompletionHandler?(statusCode: statusCode, operationError: Errors.HTTP(statusCode: statusCode, response: response))
         }
     }
     
--- a/Source/GetDocumentOperation.swift
+++ b/Source/GetDocumentOperation.swift
@@ -103,10 +103,10 @@ public class GetDocumentOperation: CouchDatabaseOperation {
             }
         } else {
             guard let data = data else {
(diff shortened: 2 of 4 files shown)


Source/DeleteDatabaseOperation.swift
```diff
@@ -76,7 +76,7 @@ public class DeleteDatabaseOperation : CouchOperation {
                 response = nil
             }
             
-            self.deleteDatabaseCompletionHandler?(statusCode: statusCode, operationError:Errors.CreateDatabaseFailed(statusCode: statusCode, jsonResponse: response))
+            self.deleteDatabaseCompletionHandler?(statusCode: statusCode, operationError:Errors.HTTP(statusCode: statusCode, response: response))
```
ricellis: Need a space in `operationError:Errors.HTTP(...`

Source/GetDocumentOperation.swift
```diff
@@ -103,10 +103,10 @@ public class GetDocumentOperation: CouchDatabaseOperation {
             }
         } else {
             guard let data = data else {
-                callCompletionHandler(error: Errors.GetDocumentFailed(statusCode: statusCode, jsonResponse: nil))
+                callCompletionHandler(error: Errors.HTTP(statusCode: statusCode, response: nil))
                 return
             }
-            callCompletionHandler(error:Errors.GetDocumentFailed(statusCode: statusCode, jsonResponse: String(data: data, encoding: NSUTF8StringEncoding )))
+            callCompletionHandler(error:Errors.HTTP(statusCode: statusCode, response: String(data: data, encoding: NSUTF8StringEncoding )))
```
ricellis: Need a space in `error:Errors.HTTP(...`

ricellis:
Minor spacing issues, which would probably get addressed by the formatter when we have one, but it is probably better if we introduce as few as possible anyway.
Otherwise +1.

Issue #43, opened 2016-04-29
Rework completion blocks


rhyshort:
Completion blocks need to be reworked to add more power to the operations. 
High level task to improve the API.

We need to do the following:

- [x]  Extend Errors Enum to Match HTTP status codes
- [x] Move to using the HTTP status Code errors
- [x] Remove non HTTP errors minus validation and invalid json.
- [x] Create a struct to contain HTTP info, such as status code and headers.
- [x] Rework completion handlers to be in the form of `((response:[String:AnyObject]?, httpInfo: HttpInfo?, error:ErrorProtocol?)-> Void)? `
- [x] Move completion handler definition to `CouchOperation`
- [x] Move processResponse logic `CouchOperation` 
- [x] Create new base class for operations which use row handlers and define the property `var ((document:[String:AnyObject])-> Void)?

Completion handler behaviour
{httpInfo} represents the HTTPinfo struct to be created in task 2

- 2xx `completionHandler?([String:AnyObject], {httpInfo} ,nil)`
- !2xx `completionHandler?([String:AnyObject], {httpInfo} ,HTTP(statusCode, responseString))`
- HTML! `completionHandler?(nil, {httpInfo} ,Unavailable(statusCode, responseString))`
- OperationCondition Failure `completionHandler?(nil, nil,InvalidJson)`

PR #48, opened 2016-05-03
Create Http info struct

rhyshort:
Part of #43 

## What 
Create HTTP info struct which will contain HTTP response information.

## Testing
None, this just defines the HTTP struct.

## Reviewers
reviewer @ricellis

### DIFF
--- a/Source/RequestExecutor.swift
+++ b/Source/RequestExecutor.swift
@@ -17,6 +17,20 @@
 
 import Foundation
 
+/**
+    Contains HTTP response information.
+ */
+public struct HttpInfo {
+    /**
+     The status code of the HTTP request.
+    */
+    let statusCode: Int
+    /**
+      The headers that were returned by the server.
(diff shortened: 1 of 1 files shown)


ricellis:
+1

Source/RequestExecutor.swift
```diff
@@ -17,6 +17,20 @@
 
 import Foundation
 
+/**
+    Contains information about a succesful HTTP request
```
ricellis: Really this is information about the HTTP response.
Also typo in successful (missing s), but I'd be happy to drop that entirely because it is ambiguous - in some places we say 2xx is successul, but this struct would contain info in 4xx cases too.
I'd probably go with "Contains HTTP response information."

rhyshort: Done in d883e19

ricellis:
+1

PR #50, opened 2016-05-05
Remove unused errors

rhyshort:
Part of #43 

Remove values from Errors enum which are no longer required, and move QueryViewOperation to using `.HTTP` error.

reviewer: @brynh

### DIFF
--- a/Source/QueryViewOperation.swift
+++ b/Source/QueryViewOperation.swift
@@ -353,7 +353,7 @@ public class QueryViewOperation: CouchDatabaseOperation {
                 callCompletionHandler(error: error)
             }
         } else {
-            callCompletionHandler(error: Errors.QueryViewFailed(statusCode: statusCode, response: String(data, NSUTF8StringEncoding)))
+            callCompletionHandler(error: Errors.HTTP(statusCode: statusCode, response: String(data, NSUTF8StringEncoding)))
         }
     }
     
--- a/Source/CouchOperation.swift
+++ b/Source/CouchOperation.swift
@@ -22,41 +22,10 @@ import Foundation
  */
 enum Errors : ErrorProtocol {
     /**
-    Creating a database failed.
-    */
-    case CreateDatabaseFailed(statusCode:Int, jsonResponse:String?)
-    /**
-    Deleting a database failed.
-    */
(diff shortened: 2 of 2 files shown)


Source/QueryViewOperation.swift
```diff
@@ -353,7 +353,7 @@ public class QueryViewOperation: CouchDatabaseOperation {
                 callCompletionHandler(error: error)
             }
         } else {
-            callCompletionHandler(error: Errors.QueryViewFailed(statusCode: statusCode, response: String(data, NSUTF8StringEncoding)))
+            callCompletionHandler(error: Errors.HTTP(statusCode: statusCode, response: String(data, NSUTF8StringEncoding)))
```
brynh: The comment on `Errors.HTTP` says it's used for a 4xx or 5xx error, whereas here it's being used for things that aren't 200 - which isn't quite the same (although a 4xx or 5xx should be the most likely errors if the server's behaving itself). I wonder if we should change the comment on `Errors.HTTP` to remove the bit about 4xx or 5xx if we're using it this way?

rhyshort: We could however we are unlikely to see a `3xx` or `1xx` response, `NSURLSession` should handle all the redirect `3xx` responses, the others we shouldn't see since Couch doesn't respond with those. When we get to the operation we should get a 2xx, 4xx or 5xx status codes so I think we are valid here. Unless my understanding of `NSURLSession` is off.

brynh: My point is really that although the server shouldn't respond with something unexpected, we can't  guarantee it won't - if for example there's a server side bug that causes it to return a 2xx that's not a 200 then we create an HTTP Error that's not a 4xx or 5xx. The comment on `Errors.HTTP` currently implies it can only be a 4xx or 4xx which wouldn't necessarily be the case here if there was a server-side bug causing an unexpected status to be returned. I'd prefer the wording on the comment for `Errors.HTTP` to be something more like `An unexpected HTTP status code (e.g. 4xx or 5xx) was received.` then because 4xx or 5x is only an example it makes more sense to me to use it like we are here.

rhyshort: done in 2782fb4

brynh:
+1

PR #51, opened 2016-05-06
Change completion handler signatures

rhyshort:
## What
Change the completion handler signatures to be of type `((response:[String:AnyObject]?, httpInfo:HttpInfo?,error:ErrorProtocol?) -> Void)?`

## How
- Change completion handler signature across all operations.
- Changes the HTTPOperation protocol to pass in `HttpInfo` instead of `statusCode`

## Testing
No new testing, all tests have been updated to reflect the completion handler changes however.

## Reviewers
reviewer @ricellis 
## Issues
Part of #43

### DIFF
--- a/Source/CreateDatabaseOperation.swift
+++ b/Source/CreateDatabaseOperation.swift
@@ -44,10 +44,11 @@ public class CreateDatabaseOperation : CouchOperation {
     /**
         A block to call when the operation completes
      
-     - parameter statusCode: the status code of the http response
-     - parameter operationError: The error that occured, or `nil` if processed succesfully.
+     - parameter response: The complete JSON response.
+     - parameter httpInfo: Information about the HTTP response.
+     - parameter error: The error that occured, or `nil` if processed succesfully.
      */
-    public var createDatabaseCompletionHandler : ((statusCode:Int?, operationError:ErrorProtocol?) -> Void)? = nil
+    public var createDatabaseCompletionHandler : ((response:[String:AnyObject]?, httpInfo: HttpInfo?, error:ErrorProtocol?)-> Void)? = nil
     
     
     public override func validate() -> Bool {
@@ -55,7 +56,7 @@ public class CreateDatabaseOperation : CouchOperation {
     }
     
     override public func callCompletionHandler(error: ErrorProtocol) {
-        self.createDatabaseCompletionHandler?(statusCode: nil, operationError: error)
+        self.createDatabaseCompletionHandler?(response:nil, httpInfo: nil, error: error)
     }
     
     public override func processResponse(data: NSData?, statusCode: Int, error: ErrorProtocol?) {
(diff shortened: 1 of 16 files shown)


Source/CreateDatabaseOperation.swift
```diff
@@ -44,42 +44,50 @@ public class CreateDatabaseOperation : CouchOperation {
...
+            let response:String?
+            if let data = data {
+               response = String(data: data, encoding: NSUTF8StringEncoding)
             } else {
                 response = nil
             }
-            
-            self.createDatabaseCompletionHandler?(statusCode:statusCode,
-                                                 operationError: Errors.HTTP(statusCode: statusCode, response: response))
+            self.createDatabaseCompletionHandler?(response: nil, httpInfo: httpInfo, error: Errors.UnexpectedJSONFormat(statusCode: httpInfo.statusCode, response: response))
```
ricellis: Should this be `response: response` instead of `nil`?

rhyshort: No, it is nil because the JSON deserialisation has failed, it doesn't match the type anymore so null is returned, and the response string is filed into the UnexpectedJSONFormat error.

ricellis: OK, I see it being used at the end of the line now, it is pretty confusing that we have so many things called response.

ricellis: As it is likely that this frequently repeated block of error checking will get moved up when there is just one place to fix it might be worth reconsidering some of the variable naming to make it a bit clearer what is really going on here.

rhyshort: Yeah I'll move it up when I move the completion handler definition up too.

Tests/TestHelpers.swift
```diff
@@ -76,11 +76,11 @@ extension XCTestCase {
...
         create.databaseName = databaseName;
-        create.createDatabaseCompletionHandler = {(statusCode, error) in
-            if let statusCode = statusCode {
-                XCTAssert(statusCode / 100 == 2)
+        create.createDatabaseCompletionHandler = {(response, httpInfo, error) in
+            if let httpInfo  = httpInfo {
+                XCTAssert(httpInfo.statusCode / 100 == 2)
             } else {
-                XCTAssertNotNil(statusCode)
+                XCTAssertNotNil(httpInfo)
```
ricellis: This will always fail in the else, maybe move it ahead of the `if let` to match the pattern elsewhere?

Tests/TestHelpers.swift
```diff
@@ -91,11 +91,11 @@ extension XCTestCase {
...
         delete.databaseName = databaseName
-        delete.deleteDatabaseCompletionHandler = {(statusCode, error) in
-            if let statusCode = statusCode {
-                XCTAssert(statusCode / 100 == 2)
+        delete.deleteDatabaseCompletionHandler = {(response, httpInfo, error) in
+            if let httpInfo = httpInfo {
+                XCTAssert(httpInfo.statusCode / 100 == 2)
             } else {
-                XCTAssertNotNil(statusCode)
+                XCTAssertNotNil(httpInfo)
```
ricellis: ditto

Tests/TestHelpers.swift
rhyshort: fixed in f526634

Tests/TestHelpers.swift
rhyshort: fixed in f526634

ricellis:
The duplication is so painful, but I know it is going away in the context of the remaining changes of #43 so +1

PR #53, opened 2016-05-09
Provide hook for executing additional handlers

rhyshort:
## What
Provide a hook into `CouchOperation` to make it possible to call additional completion handlers without reimplementing response processing logic.

## How
Added a new hook into CouchOperation processResponse(json: [String:AnyObkect]) to enable
running of additional completion handlers.

## Issues
Part of #43

### DIFF
--- a/Source/CouchOperation.swift
+++ b/Source/CouchOperation.swift
@@ -147,8 +147,8 @@ public class CouchOperation : NSOperation, HTTPRequestOperation
         do {
             if let data = data {
                 let json = try NSJSONSerialization.jsonObject(with: data) as! [String:AnyObject]
-                
                 if httpInfo.statusCode / 100 == 2 {
+                    self.processResponse(json: json)
                     self.completionHandler?(response: json, httpInfo: httpInfo, error: nil)
                 } else {
                     self.completionHandler?(response: json, httpInfo: httpInfo, error: Errors.HTTP(statusCode: httpInfo.statusCode, response: String(data: data, encoding: NSUTF8StringEncoding)))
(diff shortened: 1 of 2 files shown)


Source/CouchOperation.swift
```diff
@@ -147,7 +147,7 @@ public class CouchOperation : NSOperation, HTTPRequestOperation
         do {
             if let data = data {
                 let json = try NSJSONSerialization.jsonObject(with: data) as! [String:AnyObject]
-                
+                self.processResponse(json: json)
```
brynh: This should be in the `if httpInfo.statusCode / 100 == 2` branch.

Source/CouchOperation.swift
```diff
@@ -167,6 +167,19 @@ public class CouchOperation : NSOperation, HTTPRequestOperation
         }
         
     }
+    
+    /**
+     This method is called from the
+     `processResponse(data: NSData?, httpInfo: HttpInfo?, error: ErrorProtocol?)` method,
+     it will contain the desearlized json response in the event the request returned with a
```
brynh: Typo "deserialized"

Source/CouchOperation.swift
```diff
@@ -167,6 +167,19 @@ public class CouchOperation : NSOperation, HTTPRequestOperation
...
+     2xx status code. 
+     
+     - Note: This should be used to trigger other handlers such as a handler for
+     each row of view.
```
brynh: I think this should mention the fact it needs to be overridden - how about `- Note: This should be overridden to trigger other handlers such as a handler for each row of a returned view.`?

Source/CouchOperation.swift
rhyshort: done in 54256c0

Source/CouchOperation.swift
rhyshort: done in 54256c0

Source/CouchOperation.swift
rhyshort: fixed in 4c0c0c1

brynh:
+1

PR #58, opened 2016-05-10
Correct test assertion

### LINKED ISSUE #57: Invalid assertion following completion handler rework
Now that the completion handler is passed the error JSON response the assertion that the response is `nil` in test `testDeleteDocumentOpCompletesWithoutCallback` is incorrect.
```
Error Message
XCTAssertNil failed: "["reason": deleted, "error": not_found]" - 
Stacktrace
Tests/DeleteDocumentTests.swift:129
```

Oddly the failing test is reported as `testDeleteDocumentOpFailsValidationWhenDocIdIsMissing` in both local and Jenkins builds, but the assertion failure stack appears to be correct.
FYI @rhyshort

rhyshort:
## What
Correct test assertion which looks for a null response when the response should not be null.

## Issues

Fixes #57


ricellis:
+1

PR #59, opened 2016-05-10
Don't crash if URL generation fails.

rhyshort:
## What
Don't crash if RequestBuilder fails to generate a URL for the request.

## How
If generating a URL fails, we throw an error instead of aborting, this allows us to call the completionHandler with the error and complete the operation.


brynh:
+1

PR #60, opened 2016-05-11
Add hook for data seralisation

rhyshort:
## What
Make it possible for operations to complete the operation with error when data serialisation fails

## How
Added new `serialise` hook which is a throwing function which allows operations to error when attempting to transform the data to json.

### DIFF
--- a/Source/CouchOperation.swift
+++ b/Source/CouchOperation.swift
@@ -198,23 +198,42 @@ public class CouchOperation : NSOperation, HTTPRequestOperation
         return true
     }
     
+    /**
+     This should be used to serialise any data into the format expected by the CouchDB/Cloudant 
+     endpoint.
+     
+     - throws: An error in the event of a failure to serialise.
+     - note: This is guranteed to be  called after `validate() -> Bool` and before the 
+     `HTTPRequestOperation` properties are computed.
+    */
+    public func serialise() throws {
+        
+    }
+    
     final override public func start() {
-        // Always check for cancellation before launching the task
-        if isCancelled {
+        do {
+            // Always check for cancellation before launching the task
+            if isCancelled {
+                isFinished = true
+                return
+            }
+            
+            if !self.validate() {
+                self.callCompletionHandler(error:Errors.ValidationFailed)
+                isFinished = true
+                return
+            }
+            
+            try self.serialise()
+            
+            // start the operation
+            isExecuting = true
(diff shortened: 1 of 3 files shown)


Source/CouchOperation.swift
```diff
@@ -198,23 +198,42 @@ public class CouchOperation : NSOperation, HTTPRequestOperation
...
-            isFinished = true
-            return
-        }
+    /**
+     This should be used to serialise any data into the format expected by the CouchDB/Cloudant 
+     endpoint.
+     
+     - throws: An error in the event of a failure to serialise.
+     - note: This is guranteed to be  called after `validate() -> Bool` and before any of the 
+     properties defined in `HttpOperation` are called.
```
ricellis: > before any of the properties defined in `HttpOperation` are called.

I think this is a bit confusing. Does it really mean before any of the property values are fixed in the request? Admittedly that isn't a much better description.

Tests/QueryViewTests.swift
```diff
@@ -72,6 +72,8 @@ public class QueryViewTests : XCTestCase {
         view.stale = .Ok
         view.startKey = "startKey"
         view.databaseName = self.dbName
+        XCTAssert(view.validate())
+        try view.serialise()
         
         XCTAssert(view.validate())
```
ricellis: This is now a duplicate call of `XCTAssert(view.validate())` (for all 3 tests you modified).

rhyshort: fixed in 9f6a99a

Source/CouchOperation.swift
rhyshort: done in 1d94831

Source/CouchOperation.swift
```diff
@@ -203,8 +203,8 @@ public class CouchOperation : NSOperation, HTTPRequestOperation
      endpoint.
      
      - throws: An error in the event of a failure to serialise.
-     - note: This is guranteed to be  called after `validate() -> Bool` and before any of the 
-     properties defined in `HttpOperation` are called.
+     - note: This is guranteed to be  called after `validate() -> Bool` and before the HttpOperation
```
ricellis: was it intentional to remove the backticks from `HttpOperation`?

rhyshort: fixed in b766c1d

ricellis:
+1

PR #61, opened 2016-05-11
Move callCompletionHandler implementation to CouchOperation

rhyshort:
## What
Move callCompletionHandler implementation to CouchOperation.

## Why
Now that completionHandlers are common, there is no need for each operation to override the implementation on CouchOperation.


ricellis:
You missed the ones in the `*DatabaseOperation`s.

ricellis:
+1

PR #66, opened 2016-05-11
Add Operation to create attachments

### LINKED ISSUE #63: Create attachment operation
We need an operation to create attachments on documents.

rhyshort:
## What
Add the ability to create attachments for a document on the remote server.

## How
Created PutAttachmentOperation which contains the functionality to upload attachments to the server.

## Reviewers
reviewer @ricellis

## Issues 

Fixes #63

### DIFF
--- a/Source/PutAttachmentOperation.swift
+++ b/Source/PutAttachmentOperation.swift
@@ -47,7 +47,6 @@ public class PutAttachmentOperation: CouchDatabaseOperation {
     /**
      The id of the document that the attachment should be attached to.
      
-     - Note: mut be set for the attachment operation to successfully run.
      */
     public var docId: String?
     
@@ -73,8 +72,11 @@ public class PutAttachmentOperation: CouchDatabaseOperation {
     
     
     public override func validate() -> Bool {
-        if docId == nil{
-            return false;
+        if !super.validate() {
+            return false
+        }
+        if docId == nil {
+            return false
         }
         
         if revId == nil {
@@ -86,14 +88,14 @@ public class PutAttachmentOperation: CouchDatabaseOperation {
         }
         
         if data == nil {
-            return false;
+            return false
         }
         
         if contentType == nil {
             return false
         }
         
-        return super.validate()
+        return true
     }
     
     public override var httpMethod: String {
--- a/Source/PutAttachmentOperation.swift
+++ b/Source/PutAttachmentOperation.swift
@@ -1,0 +1,119 @@
+//
+//  PutAttachmentOperation.swift
+//  SwiftCloudant
+//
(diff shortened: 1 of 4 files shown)


Source/PutAttachmentOperation.swift
```diff
@@ -0,0 +1,119 @@
...
+        
+        if data == nil {
+            return false;
+        }
+        
+        if contentType == nil {
+            return false
+        }
+        
+        return super.validate()
```
ricellis: I think we agreed that the preferred pattern was that `super.validate` should be called first.

Source/PutAttachmentOperation.swift
```diff
@@ -0,0 +1,119 @@
...
+ database.add(operation: putAttachment)
+ 
+ 
+ */
+public class PutAttachmentOperation: CouchDatabaseOperation {
+    
+    /**
+     The id of the document that the attachment should be attached to.
+     
+     - Note: mut be set for the attachment operation to successfully run.
```
ricellis: All the properties are required to be set so it is strange to call this out on `docId` but not the others. Given that it is stated at the class level I think this can be removed or if you prefer copy it for all the properties and remove the class level comment.

Tests/PutAttachmentTests.swift
```diff
@@ -0,0 +1,200 @@
...
+        client?[dbName!].add(operation: createDoc)
+        createDoc.waitUntilFinished()
+    }
+    
+    override func tearDown() {
+        deleteDatabase(databaseName: dbName!, client: client!)
+        
+        super.tearDown()
+        
+        print("Deleted database: \(dbName!)")
```
ricellis: Is this print really necessary?

ricellis:
It would be so nice if each of the tests didn't have so much duplication.

Source/PutAttachmentOperation.swift
rhyshort: We did, I think I wrote / submitted this before the agreement will fix

Source/PutAttachmentOperation.swift
rhyshort: removed in c25f393

Source/PutAttachmentOperation.swift
rhyshort: reworked in c25f393

Tests/PutAttachmentTests.swift
rhyshort: totally :P seriously no, removed in 364f4cf

rhyshort:
Reduced duplication in 5bd2523

ricellis:
+1

PR #79, opened 2016-06-02
Add ability to read attachments from the database.

### LINKED ISSUE #64: Read attachment operation

rhyshort:
## What

Add the ability to read attachments from the database.

## How

Added the `ReadAttachmentOperation`, this has a couple of limitations,
due to the object hierarchy, the ReadAttachmentOperation defines a new completion handler,
readAttachmentCompletionHandler, to receive the data from the database. This does mean the standard
completion handler *will* not be called for ReadAttachmentOperation instances.

## Testing

New tests added, see `ReadAttachmentTests` for details.

## Issues

Resolves #64

### DIFF
--- a/Tests/ReadAttachmentTests.swift
+++ b/Tests/ReadAttachmentTests.swift
@@ -1,0 +1,139 @@
+//
+//  ReadAttachmentTests.swift
+//  SwiftCloudant
+//
+//  Created by Rhys Short on 02/06/2016.
+//  Copyright © 2016 IBM. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+
+import Foundation
+import XCTest
+@testable import SwiftCloudant
+
+class ReadAttachmentTests: XCTestCase {
+    
+    var dbName: String? = nil
+    var client: CouchDBClient? = nil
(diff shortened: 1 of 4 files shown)


Tests/ReadAttachmentTests.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+    
+    override func tearDown() {
+        deleteDatabase(databaseName: dbName!, client: client!)
+        
+        super.tearDown()
+    }
+    
+    func testReadAttachment() {
+        let expectation = self.expectation(withDescription: "read attchment")
```
brynh: Typo `attchment` -> `attachment`

Tests/ReadAttachmentTests.swift
```diff
@@ -0,0 +1,139 @@
...
+            XCTAssertNil(error)
+            XCTAssertNotNil(info)
+            if let info = info {
+                XCTAssert(info.statusCode / 100 == 2)
+            }
+            
+            XCTAssertNotNil(data)
+            if let data = data {
+                let attxt = NSString(data: data, encoding: NSUTF8StringEncoding)
+                XCTAssertEqual(self?.attachment, attxt)
```
brynh: Comparing `String?` with `NSString?` here needs a cast.

rhyshort: hmmmm, it works with the May 31 Snapshot, maybe they changed it back so the compiler auto casts again?

brynh: Ah, I wondered if that might be the case. I tried to test with the May 31 Snapshot but I can't get it to work at all (Xcode crashes on startup). I assume the Jenkins build is still using the May 9 snapshot. I guess we can leave this then, but I suppose the Jenkins build needs to be updated to the latest snapshot.

rhyshort: @brynh Yeah, I just updated the snapshot on Jenkins (today has been a fun day of machine maintenance, I'll push the fix the other comment and it should go green.

Tests/ReadAttachmentTests.swift
rhyshort: fixed in 3fd5194

brynh:
+1

PR #80, opened 2016-06-06
Protocol Orientated Operations.

rhyshort:
## What
Rework the interface to be Protocol Orientated rather than Object Orientated.

## How
- Break `CouchOperation` into two parts, `Operation` as the subclass of `NSOperation` and `CouchOperation` protocol for which operations (such as CreateDatabaseOperation) implement.
  - `CouchOperation` protocol also introduces a new hook, `prcoessResponse(data:....)` this is to enable  data to be processed differently depending  on the operation context, eg GetAttachment won't deserialise data, where as DeleteAttachment would.
- Introduce new protocols that define a `JsonOperation` and a `DataOperation` this allows defining different `completionHandler`s depending on the data that is returned from the server.
- HttpProperties have been renamed, to remove the `Operation` from http naming.

## Testing

No additional tests, but all currents tests work as expected.


CHANGELOG.md
```diff
@@ -2,6 +2,27 @@
...
 - [NEW] New FindDocumentsOperation API.
 - [NEW] New DeleteAttachmentOperation API.
+- [BREAKING] `Database` class has been removed
+- [BREAKING] `CouchOperation` is now a protocol, in order to use `NSOperation` APIs
+  you need to use the `Operation` class in conjunction with objects that conform
+  to `CouchOperation`
+- [BREAKING] `CouchDatabaseOperation` is now a protocol.
+- [NEW] `JsonOperation` and `DataOperation` protocols.
+- [BREAKING] `completionHandler` property is no longer defined on `CouchOperation`
+   it's replacement is `completionHander` defined on `JsonOperation` and `DataOperation`
```
brynh: Typo `it's` -> `its`

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,229 @@
...
+    var parameters: [String: String] { get }
+    
+    /**
+     The data to send to the CouchDB endpoint.
+     */
+    var data: NSData? { get }
+    
+    
+    /**
+     The content type of the data to send to the CouchDB endpoint. This is guranteed to be called
```
brynh: Typo `guranteed` -> `guaranteed`

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,229 @@
...
+     if and only if `data` is not `nil`
+     */
+    var contentType: String { get }
+    
+    /**
+     Calls the completionHandler for this operation.
+     
+     - parameter response: The response recceived from the server, this should be on of the following types,
```
brynh: Typos:
`recceived` -> `received`
`on` -> `one`

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,229 @@
...
+     */
+    func processResponse(json: Any)
+    
+    
+    /**
+     Processes the response from CouchDB. This is **required** to call the following 
+     methods in the lifecycle of the function.
+     
+     * `processResponse(json:)` to allow an operation to provide addtional processing of the json
+        result when 2xx response code was recieved, and if the response is a JSON response.
```
brynh: Typo `recieved` -> `received`

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,229 @@
...
+     use the result of the super class implementation.
+     */
+    func validate() -> Bool
+    
+    /**
+     This should be used to serialise any data into the format expected by the CouchDB/Cloudant
+     endpoint.
+     
+     - throws: An error in the event of a failure to serialise.
+     - note: This is guranteed to be  called after `validate() -> Bool` and before the
```
brynh: Typo `guranteed` -> `guaranteed`

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,229 @@
...
+
+/** 
+ Marks an operation as a Data Only operaton. It will provide the data directly from the database.
+ */
+public protocol DataOperation: CouchOperation {
+    
+    /**
+     Sets a completion handler to run when the operation completes.
+     
+     - parameter response: - The full data reccieved from the server.
```
brynh: Typo `reccieved` -> `received`

Source/FindDocumentsOperation.swift
```diff
@@ -283,11 +288,18 @@ public class FindDocumentsOperation: CouchDatabaseOperation, MangoOperation {
         }
     }
 
-    public override func processResponse(json: [String: AnyObject]) {
-        if let docs = json["docs"] as? [[String: AnyObject]] { // Array of [String:AnyObject]
+    public  func processResponse(json: Any) {
+        if let json = json as? [String: AnyObject],
+           let docs = json["docs"] as? [[String: AnyObject]] { // Array of [String:AnyObject]
             for doc: [String: AnyObject] in docs {
                 self.documentFoundHanlder?(document: doc)
```
brynh: Typo `documentFoundHanlder` -> `documentFoundHandler`

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,229 @@
...
+     a returned view.
+     */
+    func processResponse(json: Any)
+    
+    
+    /**
+     Processes the response from CouchDB. This is **required** to call the following 
+     methods in the lifecycle of the function.
+     
+     * `processResponse(json:)` to allow an operation to provide addtional processing of the json
```
brynh: I find this a little unclear, in that the text before the list says these are required, but then they are dependent on circumstances that for `processResponse` may be impossible (if the response is not JSON). I wonder about a slight rewording making it:

> Processes the response from CouchDB. This is **required** to call the following
> methods in the lifecycle of the function when the given circumstances arise:
> * `callCompletionHandler(error:)` when an error has occurred.
> * `callCompletionHandler(response:httpInfo:error:)` when the processing of the response has completed.
> * `processResponse(json:)` when a 2xx response code was received **and** the response is a JSON response. This allows an operation to provide additional processing of the JSON.

What do you think?

Source/OperationProtocols.swift
rhyshort: fixed in 6e6b0dd

Source/OperationProtocols.swift
rhyshort: fixed in 92fd0a2

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,229 @@
...
+                response = nil
+            }
+            let error = Errors.UnexpectedJSONFormat(statusCode: httpInfo.statusCode, response: response)
+            self.callCompletionHandler(response: nil, httpInfo: httpInfo, error: error)
+        }
+    }
+}
+
+/** 
+ Marks an operation as a Data Only operaton. It will provide the data directly from the database.
```
rhyshort: fixed in 92fd0a2

Source/OperationProtocols.swift
rhyshort: fixed in 92fd0a2

Source/OperationProtocols.swift
rhyshort: fixed in 92fd0a2

Source/FindDocumentsOperation.swift
rhyshort: fixed in 588ecb4

Source/OperationProtocols.swift
rhyshort: I agree that is better, I think that one change should be made however, I think `callCompletionHandler(error:)` should read `callCompletionHandler(error:) when the processing of the response has completed with an error.` to make it clearer processing should stop,  thoughts?

brynh: Agreed, sounds good to me.

Source/OperationProtocols.swift
brynh: This hasn't been fixed yet.

Source/OperationProtocols.swift
rhyshort: done in cc64227

Source/CreateDatabaseOperation.swift
```diff
@@ -19,30 +19,32 @@ import Foundation
...
+     public var method: String {
         get {
             return "PUT"
         }
     }
 
-    public override var httpPath: String {
+    public  var endpoint: String {
         get {
             // Safe to foce unwrap validation would fail if this is nil
```
brynh: Suggest making this `Safe to force unwrap, validation would fail if this is nil` - i.e. correcting typo and adding a comma to make it easier to read.

Source/DeleteDatabaseOperation.swift
```diff
@@ -19,30 +19,32 @@ import Foundation
...
+     public var method: String {
         get {
             return "DELETE"
         }
     }
 
-    override public var httpPath: String {
+     public var endpoint: String {
         get {
             // Safe to foce unwrap validation would fail if this is nil
```
brynh: Suggest making this `Safe to force unwrap, validation would fail if this is nil` - i.e. correcting typo and adding a comma to make it easier to read.

Source/OperationProtocols.swift
rhyshort: hmm must have did a different line, fixed in de9709d

Source/CreateDatabaseOperation.swift
rhyshort: fixed in a82fa03

Source/DeleteDatabaseOperation.swift
rhyshort: fixed in 84c87b1

CHANGELOG.md
```diff
@@ -2,6 +2,27 @@
...
+| Previous Name | New Name |
+|---------------|---------|
+| `httpPath` | `endpoint`|
+| `httpRequestData` | `data` |
+| `httpContentType` | `contentType`|
+|`httpMethod` | `method`|
+
+- [BREAKING] `QueryItems` property has been removed.
+- [NEW] `parameters` property API, used to define the parameters for a request,
+for example `["rev": "1-anexamplerevisionid"]`
```
ricellis: I don't think this example is adding anything. If it is to be an example I think it needs to show the actual API usage, but as a computed property that is probably too long for the change log. Perhaps it is sufficient to change the description to something like:
>- [NEW] `parameters` property API, used to define HTTP query parameters for a CouchOperation request.

brynh:
Overall I like the move towards a more protocol based structure. However, I have a real dislike of code duplication, so I'm not keen on the duplication in the various `validate` methods. I guess short of adding some abstract classes whose abstractness isn't enforced by the compiler, there's not a simple solution to removing this duplication.

Source/CreateDatabaseOperation.swift
```diff
@@ -19,30 +19,32 @@ import Foundation
...
-            // Safe to foce unwrap validation would fail if this is nil
+            // Safe to force unwrap, validation would fail if this is nil
             return "/\(self.databaseName!)"
         }
     }
 
-    public override func validate() -> Bool {
-        return super.validate() && self.databaseName != nil // should work iirc
+    public  func validate() -> Bool {
+        return  self.databaseName != nil // should work iirc
```
ricellis: this comment isn't much use, I'm pretty sure that check works

Source/DeleteAttachmentOperation.swift
```diff
@@ -28,21 +28,69 @@ import Foundation
...
-
-    public override var httpMethod: String {
+public class DeleteAttachmentOperation: CouchDatabaseOperation, JsonOperation {
+    
+    public var databaseName: String?
+    
+    public var completionHandler: ((response: [String : AnyObject]?, httpInfo: HttpInfo?, error: ErrorProtocol?) -> Void)?
+    
+    /**
+     The id of the document that the attachment should be attached to.
```
ricellis: Since this is a deletion, this should probably read "is attached to".

Source/DeleteAttachmentOperation.swift
```diff
@@ -28,21 +28,69 @@ import Foundation
...
+     
+     */
+    public var docId: String?
+    
+    /**
+     The revision of the document that the attachment should be attached to.
```
ricellis: Again since this is a deletion, this should probably read "is attached to".

Source/DeleteAttachmentOperation.swift
```diff
@@ -28,21 +28,69 @@ import Foundation
...
+    public var parameters: [String: String] {
+        return ["rev": revId!]
+    }
+    
+    public var method: String {
         return "DELETE"
     }
-
+    
 }
```
ricellis: Newline?

Source/GetAllDatabasesOperation.swift
```diff
@@ -48,47 +43,22 @@ public class GetAllDatabasesOperation : CouchOperation {
...
+        if let json = json as? [String] {
+            for dbName in json {
+                self.databaseHandler?(databaseName: dbName)
             }
-            self.completionHandler?(response: nil, httpInfo: httpInfo, error: Errors.UnexpectedJSONFormat(statusCode: httpInfo.statusCode, response: response))
         }
     }
     
-    
 }
```
ricellis: new line

Source/CreateQueryIndexOperation.swift
```diff
@@ -215,20 +225,20 @@ public class CreateTextQueryIndexOperation: CouchDatabaseOperation, MangoOperati
...
     
-    public override var httpPath: String {
+    public var endpoint: String {
         return "/\(self.databaseName!)/_index"
     }
     
-    public override func validate() -> Bool {
-        if !super.validate() {
+    public func validate() -> Bool {
+        if databaseName == nil {
```
ricellis: Most other places seem to use the pattern `return databaseName != nil`.
If we are going to have to duplicate so much of this validation code now, it would be nice to maintain some consistency as at least we will have an idea of what to expect in each case.

Source/DeleteAttachmentOperation.swift
```diff
@@ -28,21 +28,69 @@ import Foundation
...
+     */
+    public var revId: String?
+    
+    /**
+     The name of the attachment.
+     */
+    public var attachmentName: String?
+    
+    public func validate() -> Bool {
+        if self.databaseName == nil {
```
ricellis: Again most other places have gone for the `!=nil && ...`
I think it would be good if this was updated to match.

Source/Operation.swift
```diff
@@ -0,0 +1,177 @@
+//
+//  CouchOperation.swift
```
ricellis: `Operation.swift` ?

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,228 @@
...
+
+import Foundation
+
+/**
+ A protocol that denotes an operation which can be run agasint a CouchDB instance.
+ */
+public protocol CouchOperation {
+    
+    /**
+     The CouchDB endpoint to call, for example `/exampleDB/document1/` or `/_all_dbs`
```
ricellis: Given that these endpoints match what are described in the CouchDB or Cloudant API references, perhaps we should call them "CouchDB API endpoint" in the comment?

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,228 @@
...
+     */
+    var endpoint: String { get }
+    
+    /**
+     The method to use when making the API call.
```
ricellis: I think this description should say "The HTTP method..." to help guide the usage a little further.

Source/OperationProtocols.swift
```diff
@@ -0,0 +1,228 @@
...
+     */
+    var method: String { get }
+    
+    /**
+     The query items to use for the request
```
ricellis: "query items" seems a bit strange as terminology here, perhaps "dictionary of query parameter field:value items"? I'm not sure that is much better, but query items seems so vague especially when we have Cloudant query stuff floating around as well.

rhyshort:
@brynh Unfortunately not, it is one of the downsides with this approach, I was thinking we could have the debate about weather we keep the whole validate thing or not. With the proposal that if we drop validate the bits we know need to be present eg `databaseName` for `CreateDocumentOperations` could become `init` args. (although thats something for another time).

CHANGELOG.md
rhyshort: done in 408616f

Source/CreateDatabaseOperation.swift
rhyshort: ah yes, hang over from me learning to swift :)

rhyshort: removed in 235c7d8

Source/DeleteAttachmentOperation.swift
rhyshort: done in dd9cb67

Source/DeleteAttachmentOperation.swift
rhyshort: doen in 178c205

Source/DeleteAttachmentOperation.swift
rhyshort: added in fa7304c

Source/GetAllDatabasesOperation.swift
rhyshort: done in bd2b720

Source/CreateQueryIndexOperation.swift
rhyshort: That won't work here, since we still need to unwrap the `selector` property and validate that is it valid `json`.

Source/DeleteAttachmentOperation.swift
rhyshort: I think doing the `&&` approach would make the code less readable.

Source/Operation.swift
rhyshort: updated in 9245279

Source/OperationProtocols.swift
rhyshort: done in 83de5fa

Source/OperationProtocols.swift
rhyshort: done in 84159c1

Source/OperationProtocols.swift
rhyshort: Yeah I forgot about the comment when making the rename, Ive changed the comment to `     Dictionary of query parameters, `Key` is the parameter name and `Value` is the 
     value to use for that parameter. E.g. ["rev": "1-revisionhash"].` in 97b74e0. Thoughts?

rhyshort:
@ricellis pinging you so https://github.com/cloudant/swift-cloudant/pull/80#discussion_r66811499 doesn't get lost with the collapsing of diff comments.

Source/CreateQueryIndexOperation.swift
ricellis: True enough, the diff had compressed that exact bit of functionality away so I missed it, perhaps then we should always use the opposite approach even if it is more code. As you know I agree with @brynh I generally like the protocol approach, but I am troubled by the duplication which really is one in the face of DRY.

Source/PutDocumentOperation.swift
```diff
@@ -16,7 +16,10 @@
 
 import Foundation
 
-public class PutDocumentOperation: CouchDatabaseOperation {
+public class PutDocumentOperation: CouchDatabaseOperation, JsonOperation {
```
ricellis: Looks like this is missing an example. Probably not the only place, we should tackle those in another issue.

ricellis:
There are a bunch of places where there are two spaces between `public` and a `func` where the `override` keyword has been removed. I know we're not worrying too much about formatting until we have a useful too, but I thought I'd mention it anyway.

Source/OperationProtocols.swift
ricellis: Yep, I'm happy with that.

Source/CreateQueryIndexOperation.swift
rhyshort: Yeah I don't like it either, I think we should have the discussion on what we should validate, if anything. And then go from there, like if we decide that we should remove the validation the stuff that is required can go into the `init` method as being required rather than optional properties.

Source/DeleteAttachmentOperation.swift
ricellis: Given that some places require us to do more work that prevents using the `return cond1 && cond2` pattern, perhaps we should just explicitly expand everywhere? I guess we can move all that into a separate validate discussion.

Source/PutDocumentOperation.swift
rhyshort: issue #81 opened to over this issue.

Source/DeleteAttachmentOperation.swift
rhyshort: Agreed, lets do it as part of the validation discussion and make changes from the outcome of that.

ricellis:
+1

brynh:
+1

PR #84, opened 2016-06-17
Use Swift Package manager for Testing

rhyshort:
## What

Use the Swift Package manager for testing.

## How

- Remove the dependancy  on CocoaPods and OHHTTPStubs for testing.
- Removed Xcode project, a new one can be generated using the `swift package --generate-xcodeproj` command.
- Tests which used stubbing now use a lib dispatch to simulate an operation being executed.
- New JSONResponse which is an Array and String literal convertible so it can be used in a similar way to OHTTPStubs.

- For `SessionCookieInterceptor` custom `URLProtocol`s exist to enable the request to process, as a result a test only `init` method exists to inject the session which contains the `URLProtocol`s.

- Unfortunately this removes the ability to configure tests they will continue run using the default of `localhost:5984`

## Testing

All tests pass.


ricellis:
If the xcode proj is going to be generated shouldn't the git ignore be updated to ignore the xcode files?

Tests/InterceptorTests.swift
```diff
@@ -83,3 +65,67 @@ class InterceptorTests: XCTestCase {
...
     }
 }
+
+class CookieSessionHTTPURLProtocol: NSURLProtocol {
+    
+    override class func canonicalRequest(for request: NSURLRequest) -> NSURLRequest {
+        return request
+    }
+    
+    func sendResonse(statusCode: Int, headers: [String:String] = [:], json: JSONResponse){
```
ricellis: sendRes**p**onse

Tests/InterceptorTests.swift
```diff
@@ -83,3 +65,67 @@ class InterceptorTests: XCTestCase {
...
+            self.sendResonse(statusCode: statusCode, headers: headers, data: data)
+        } catch {
+            
+            NSLog("Failed to create NSData for JSONResponse, error:\(error)")
+            self.sendResonse(statusCode: statusCode, headers: headers, data: NSData())
+            
+        }
+    }
+    
+    func sendResonse(statusCode: Int, headers: [String:String] = [:], data: NSData){
```
ricellis: sendRes**p**onse again

rhyshort:
@ricellis added new ignored in c96ab5b

ricellis:
+1 once those minor fixes are addressed.
As I said I'm concerned about the simulation requiring us to duplicate what happens in an execute method, but I appreciate this is a stop gap until we can get a more Swift3 compatible solution in the form of mocks or a mock server for example.

rhyshort:
I agree, #85, will track the code duplication so we can fix it.

Issue #87, opened 2016-06-18
HttpInfo has incorrect property ACL.


rhyshort:
- SwiftCloudant version(s): Head.
- Swift version(s): 3 Preview 1
- Platform(s): OSX
- A small code sample that demonstrates the issue.

```swift
/**
 Contains HTTP response information.
 */
public struct HttpInfo {
    /**
     The status code of the HTTP request.
     */
    let statusCode: Int 
    /**
     The headers that were returned by the server.
     */
    let headers: [String: String]
}

```

The property definitions need to have to be `public`

Issue #86, opened 2016-06-18
PutDocumentOperation initialiser is inaccessible 


rhyshort:
- SwiftCloudant version(s): `HEAD` of master, (with modifications for swift 3 preview 1).
- Swift version(s):  Swift 3 Preview 1
- A small code sample that demonstrates the issue:

```swift
import SwiftCloudant

let put = PutDocumentOperation()
```

rhyshort:
Chances are this will be a problem across other classes in the Library. This isn't a problem our test cases would pick up due to use of the `@testable import` statement.

rhyshort:
Confirmed that `QueryViewOperation` is also effected.

PR #90, opened 2016-06-28
Correct Access control specifiers

rhyshort:
## What

Correct the access control specifiers for init methods and `HttpInfo`

## How

- Create public init methods for operations
- Make properties on HttpInfo public.

## Testing

None, test code isn't effected due to the `@testable` attribute for the imports.
## Issues

#86 
#87


ricellis:
+1

PR #91, opened 2016-06-28
Add LinuxMain file to run tests on linux.

### LINKED ISSUE #70: Add LinuxMain to Tests
To run Tests on linux, the file LinuxMain needs to be present, with defines the contents of the tests.

rhyshort:
Add LinuxMain.swift to enable running tests on linux.

Fixes #70

### DIFF
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -1,0 +1,22 @@
+//
+//  LinuxMain.swift
+//  SwiftCloudant
+//
+//  Created by Rhys Short on 28/06/2016.
+//  Copyright © 2016 IBM Corp.
+//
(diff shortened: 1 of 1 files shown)


Tests/LinuxMain.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  LinuxMain.swift
+//  SwiftCloudant
+//
+//  Created by Rhys Short on 28/06/2016.
+//  Copyright (c) 2016 IBM Corp.
```
ricellis: Should be © if the file is unicode or otherwise (C).

ricellis:
+1 besides the copyright symbol

Issue #73, opened 2016-05-20
Rename HttpInfo to HTTPInfo


rhyshort:
The `HttpInfo` struct should follow naming convention and have `HTTP` capitalised.

PR #92, opened 2016-06-28
Rename HttpInfo to HTTPInfo

rhyshort:
## What

Rename `HttpInfo` to `HTTPInfo`

## How
Rename `HttpInfo` to `HTTPInfo`

## Why 
Making the `HTTP` upper case in `HTTPInfo` makes it more in keeping with the swift API guidelines  

## Issues

#73


ricellis:
+1

PR #93, opened 2016-06-29
Linux Support

rhyshort:
## What

Add initial linux support.

## How

- Add conditional compilation sections to get around the difference in bridging on Linux and Apple Platforms.

## Testing

No test changes, until we can compile the library on Linux (this requires more work in Swift Foundation), tests cannot be compiled to see the differences.

### DIFF
--- a/Source/CreateQueryIndexOperation.swift
+++ b/Source/CreateQueryIndexOperation.swift
@@ -93,25 +93,46 @@ public class CreateJsonQueryIndexOperation: CouchDatabaseOperation, MangoOperati
     }
 
     public func serialise() throws {
-
-        var jsonDict: [String:AnyObject] = ["type": "json"]
+        
+        #if os(Linux)
+            var jsonDict: [String:AnyObject] = ["type": "json".bridge()]
+        #else
+            var jsonDict: [String:AnyObject] = ["type": "json"]
+        #endif
 
         if let fields = fields {
             var index: [String: AnyObject] = [:]
-            index["fields"] = transform(sortArray: fields) as NSArray
-            jsonDict["index"] = index as NSDictionary
+            #if os(Linux)
+                index["fields"] = transform(sortArray: fields).bridge()
+                jsonDict["index"] = index.bridge()
+            #else
+                index["fields"] = transform(sortArray: fields) as NSArray
+                jsonDict["index"] = index as NSDictionary
+            #endif
+            
         }
 
         if let indexName = indexName {
-            jsonDict["name"] = indexName as NSString
+            #if os(Linux)
+                jsonDict["name"] = indexName.bridge()
+            #else
+                jsonDict["name"] = indexName as NSString
+            #endif
         }
 
         if let designDoc = designDoc {
-            jsonDict["ddoc"] = designDoc as NSString
+            #if os(Linux)
+                jsonDict["ddoc"] = designDoc.bridge()
+            #else
+                jsonDict["ddoc"] = designDoc as NSString
+            #endif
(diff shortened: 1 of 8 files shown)


Source/CreateQueryIndexOperation.swift
```diff
@@ -259,41 +284,82 @@ public class CreateTextQueryIndexOperation: CouchDatabaseOperation, MangoOperati
...
-            }
+            #if os(Linux)
+                jsonDict["type"] = "text".bridge()
+                
+                if let indexName = indexName {
+                    jsonDict["name"] = indexName.bridge()
+                }
+                
+                if let fields = fields {
+                    index["fields"] = transform(fields: fields)
```
ricellis: since the `transform` func calls returns using `as NSArray` do we need an additional conditional compilation there or to call `.bridge()` here?

rhyshort: No, the `transform` func should call `bridge` to return the NSArray.

ricellis: Ha, when I looked it wasn't there, I must have had the wrong version of transform loaded.

Source/Operation.swift
```diff
@@ -102,7 +102,7 @@ public class Operation: NSOperation, HTTPRequestOperation
         }
     }
 
-    internal var rootURL: NSURL = NSURL()
+    internal var rootURL: NSURL = NSURL(string: "http://example.cloudant.com")!
```
ricellis: Since this is a real thing I suggest we use something like `http://cloudant.placeholder.invalid`
https://tools.ietf.org/html/rfc2606#section-2
https://tools.ietf.org/html/rfc6761#section-6.4

Source/SessionCookieInterceptor.swift
```diff
@@ -66,13 +66,21 @@ public class SessionCookieInterceptor: HTTPInterceptor
...
-            ctx.shouldRetry = true
-            self.cookie = nil
-        } else if let cookieHeader = response.allHeaderFields["Set-Cookie"] as? String {
-            cookie = cookieHeader.components(separatedBy: ";").first
-        }
+        #if os(Linux)
+            if response.statusCode == 403 || response.statusCode == 401 {
+                ctx.shouldRetry = true
+                self.cookie = nil
+            } else if let cookieHeader = response.allHeaderFields["Set-Cookie"] as? String {
```
ricellis: Shouldn't this bridge? It looks identical to the `else` case?

rhyshort: ah yes I forgot about this, not actually sure what to do the compiler kicks out ``` error: downcast from 'String?' to 'String' only unwraps optionals; did you mean to use '!'?
            } else if let cookieHeader = response.allHeaderFields["Set-Cookie"] as? String {
```

Source/Operation.swift
rhyshort: done in dff5996

Source/SessionCookieInterceptor.swift
rhyshort: Hmm I think we probably should ignore this for now, looks like a compiler error. 

`response.allHeaderFields["Set-Cookie"] as? String` should be evaluated as an optional cast, with the `if let cookieHeader` part unwrapping the optional if the value is present. thoughts @ricellis?

ricellis: OK, leave this one out and hopefully we'll be able to resolve it with a future update.

ricellis:
+1

PR #95, opened 2016-06-30
Support for the 2016-06-30 Swift 3 Snapshot

### LINKED ISSUE #82: Swift 3 Preview 1 Changes: NS Prefix removal.
There are a bunch of changes to the Swift Runtime in the Swift Preview 1. 

A number of Swift Evolution proposals have landed for OS X. These include new Objc Bridgeable types. These types should replace their NS Prefixed version. [SE-0086](https://github.com/apple/swift-evolution/blob/master/proposals/0086-drop-foundation-ns.md)

Examples

`NSData` -> `Data`
`NSURLSession` -> `URLSession`
`NSJSONSeralisation` -> `JSONSeralisation`

In addition to this `NSMutableData` has been replace with `Data` as a value type see [SE-0069](https://github.com/apple/swift-evolution/blob/master/proposals/0069-swift-mutability-for-foundation.md).

### LINKED ISSUE #83: Swift Preview 1 Changes - Dispatch
LibDispatch has been updated to provide more object like interfaces to the C Library, `SessionCookieInterceptor` makes a lot of use of lib dispatch to control flow while waiting for the cookie to be returned from the server. This class will need to use the swiftized APIs. See [SE-0088](https://github.com/apple/swift-evolution/blob/master/proposals/0088-libdispatch-for-swift3.md)

rhyshort:
## What

Support for the 2016-06-30 Swift 3 Snapshot

## How

- Remove NS Prefixes from some classes, converting them to value types.
- Use modern Dispatch
   - This changes means we need to wait forever for the cookie task to complete before continuing.

## Testing

Tests also updated, they currently all pass

## Issues

Fixes #83 
Fixes #82

### DIFF
--- a/Source/SessionCookieInterceptor.swift
+++ b/Source/SessionCookieInterceptor.swift
@@ -29,7 +29,7 @@ public class SessionCookieInterceptor: HTTPInterceptor
     /**
      The requestBody to use when performing HTTP requests to the `_session` endpoint.
      */
-    let sessionRequestBody: NSData
+    let sessionRequestBody: Data
     /**
      Determines if the `_session` request should be made
      */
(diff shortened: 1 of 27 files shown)


Source/SessionCookieInterceptor.swift
```diff
@@ -181,7 +177,8 @@ public class SessionCookieInterceptor: HTTPInterceptor
             }
         }
         task.resume()
-        dispatch_semaphore_wait(semaphore, dispatch_time(DISPATCH_TIME_NOW, self.sessionCookieTimeout * Int64(NSEC_PER_SEC)))
+//        semaphore.wait(timeout: DispatchTime.now(dispatch_time_t(DispatchTime.now, self.sessionCookieTimeout * Int64(NSEC_PER_SEC)))
```
ricellis: Either add a TODO about how this will be added back or delete it.

ricellis:
+1

Source/SessionCookieInterceptor.swift
rhyshort: done, see commit 317dc41

PR #101, opened 2016-07-08
Correct access level for `Sort` struct members

### LINKED ISSUE #98: Sort struct in inaccessible for CreateJsonQueryIndexOperation

- SwiftCloudant version(s) that are affected by this issue. 0.3
- Swift version(s) Swift-Dev-Snapshot-06-06
- Platform(s) All

Creating a Sort object does not work, giving the following error:

> Sort initializer is inaccessible due to internal protection level

```
let sort = Sort(field:"_id", sort: SortDirection.Desc)
```

I need this to work so I can assign it to the `fields` parameter of `CreateJsonQueryIndexOperation`
To replicate, try creating a Sort object in an app that uses Swift-Cloudant as a package.

rhyshort:
## What

Correct access level for `Sort` struct members, it should be directly defined as `public` rather than have
no access modifier.

## Issues

Fixes #98

### DIFF
--- a/Source/FindDocumentsOperation.swift
+++ b/Source/FindDocumentsOperation.swift
@@ -38,8 +38,12 @@ public struct Sort {
     /**
      The field on which to sort
      */
-    let field: String
-    let sort: SortDirection?
+    public let field: String
+    
+    /**
+     The direction in which to sort.
+     */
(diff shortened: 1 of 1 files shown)


ricellis:
+1

Issue #96, opened 2016-07-06
Follow API naming Guidelines 


rhyshort:
Some APIs do not follow the API naming Guidelines for Swift 3. Such as the enum `Stale` which has the elements `Ok` and `UpdateAfter` in Swift 3 they should be named `ok` and `updateAfter` respectively. There are probably other APIs which do not match the convention, we need to audit and propose replacements for each API that does not follow the guidelines.

The guidelines can be found [here](https://swift.org/documentation/api-design-guidelines/).

rhyshort:
# Swift API Guidlines and SwiftCloudant

Along with Swift 3, there are official [API guidelines](https://swift.org/documentation/api-design-guidelines/) approved by the Swift community community via the swift evolution process.

SwiftCloudant currently does not follow the guidelines that are laid out by the community. The most common violation of the guidelines are enum capitalisation, for example, `.Ok` should be `.ok`.
This leads to our APIs being difficult to work with and do not behave in the manor that the user would expect.

Our aim should be to make the API as expressive as possible and behave has expected. This may mean departing form the CouchDB way of naming parameters to make their purpose clearer. The example of that
would be `updateSeq`, which determines if the last update sequence number should be included in the response. It does not provide a sequence number to be used in the request.


## General

>> Changes discussed here will affect multiple classes, protocols or extensions.

There are some potential issues with the API, with Swift's explicit nullablity constructs, it is odd that we allow the the assigning of `nil` to options which are required. For example
`docId` is always required for operations operating on documents or on attachments.

To simplify documentation, validation logic and behave as the user would expect, should required parameters become required for initialization and become explicitly non-null? This goes beyond
just document IDs and covers each of the parameters which are required as per the validation rules for the operation. `databaseName` should be made non-optional since all 
operations which operate on databases will require the name of the database to be set.


The naming of parameters are abbreviated, perhaps the abbreviations should be changed to make more sense. For example the `docId` should at least be `documentID` or `docID`. However
with some operations such as `PutDocumentOperation` using just `id` would have make sense, especially since we currently use `body` to refer to the document body. It would also tie 
the parameters more closely to the context they appear in. However it does have the potential to cause confusion with the different naming on different operations. For example:

```swift

public class GetDocumentOperation : CouchOperation, CouchDatabaseOperation, JSONOperation {
	
	// ...

	public init(id: String) {
		// ....
	}

	// ...
}

[...]

PR #109, opened 2016-07-20
Follow API Guidelines 

rhyshort:
## What

Changed APIs to follow the swift API guidelines.

## How

See issue #96 for full list of transformations.

## Testing

No new tests, no longer applicable tests have been removed.

## Issues

Part of #96.

## Notes

`CouchDBClient` methods haven't been touched in this PR, I'm still unsure about the names that I proposed in the issue.
The code base will need another look through to make sure that the operations are consistent with naming and parameter ordering etc. before #96 can be resolved.

### DIFF
--- a/Source/CreateDatabaseOperation.swift
+++ b/Source/CreateDatabaseOperation.swift
@@ -19,18 +19,25 @@ import Foundation
 /**
  An operation to create a database in a CouchDB instance.
  */
-public class CreateDatabaseOperation: CouchOperation, JsonOperation {
+public class CreateDatabaseOperation: CouchOperation, JSONOperation {
 
-    public init() { }
+    /**
+     Creates the operation.
+     
+     - parameter name: The name of the database this operation will create.
+     - parameter completionHandler: optional handler to call when the operation completes.
+    */
+    public init(name: String, completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: ErrorProtocol?) -> Void)? = nil) {
+        self.name = name
+        self.completionHandler = completionHandler
+    }
     
-    public var completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: ErrorProtocol?) -> Void)?
+    public let completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: ErrorProtocol?) -> Void)?
     
     /**
      The name of the database to create.
-
-     This is required to be set before the operation can execute succesfully.
      */
-    public var databaseName: String?
+    public let name: String
 
      public var method: String {
         get {
@@ -40,13 +47,8 @@ public class CreateDatabaseOperation: CouchOperation, JsonOperation {
 
     public var endpoint: String {
         get {
-            // Safe to force unwrap, validation would fail if this is nil
-            return "/\(self.databaseName!)"
+            return "/\(self.name)"
         }
-    }
-
-    public func validate() -> Bool {
-        return  self.databaseName != nil
     }
 
 }
--- a/Source/CreateQueryIndexOperation.swift
+++ b/Source/CreateQueryIndexOperation.swift
@@ -22,12 +22,10 @@ import Foundation
  
     Usage example:
     ```
-    let index = CreateJsonQueryIndexOperation()
-    index.indexName = "exampleIndex"
-    index.fields = [Sort(field:"food", sort: .Desc)]
-    index.designDoc = "examples"
-    index.databaseName = "exampledb"
-    index.completionHandler = { (response, httpInfo, error) in
+    let index = CreateJSONQueryIndexOperation(databaseName: "exampledb",
+        designDocumentID: "examples",
+        name:"exampleIndex", 
+        fields: [Sort(field:"food", sort: .desc)]) { (response, httpInfo, error) in
         if let error = error {
             // handle the error
         } else {
@@ -39,35 +37,47 @@ import Foundation
     client.add(operation: index)
     ```
  */
-public class CreateJsonQueryIndexOperation: CouchDatabaseOperation, MangoOperation, JsonOperation {
+public class CreateJSONQueryIndexOperation: CouchDatabaseOperation, MangoOperation, JSONOperation {
     
-    public init() { }
+    /**
+     Creates the operation
+     - parameter databaseName : The name of the database where the index should be created.
+     - parameter designDocumentID : The ID of the design document where the index should be saved, 
+     if set to `nil` the server will create a new design document with a generated ID.
+     - parameter fields : the fields to be indexed.
+     - parameter completionHandler: block to run when the operation completes.
+    */
+    public init(databaseName: String,
+            designDocumentID: String? = nil,
+                        name: String? = nil,
+                      fields: [Sort],
+                              completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: ErrorProtocol?) -> Void)? = nil) {
+        self.databaseName = databaseName
+        self.fields = fields
+        self.designDocumentID = designDocumentID
+        self.name = name
+        self.completionHandler = completionHandler
+    }
     
-    public var databaseName: String?
+    public let databaseName: String
     
-    public var completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: ErrorProtocol?) -> Void)?
+    public let completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: ErrorProtocol?) -> Void)?
     
     /**
      The name of the index.
-     - Note: Optional, if this parameter is not set, it will be omitted from the request
-     and the server default will apply.
      */
-    public var indexName: String?
+    public let name: String?
     
     /**
      The fields to which the index will be applied.
-     
-     - Note: Required, this parameter needs to be set for the operation to successfully complete.
      */
-    public var fields: [Sort]?
+    public let fields: [Sort]
     
     /**
-     The name of the design doc this index should be saved to.
-     
-     - Note: Optional, if this parameter is not set, it will be omitted from the request
-     and the server default will apply.
+     The ID of the design document that this index should be saved to. If `nil` the server will
+     create a new design document with a generated ID.
      */
-    public var designDoc: String?
+    public let designDocumentID: String?
 
     private var jsonData: Data?
     
@@ -81,15 +91,7 @@ public class CreateJsonQueryIndexOperation: CouchDatabaseOperation, MangoOperati
     }
     
     public var endpoint: String {
-        return "/\(self.databaseName!)/_index"
-    }
-    
-    public func validate() -> Bool {
-        if databaseName == nil {
-            return false
-        }
-        
-        return fields != nil
+        return "/\(self.databaseName)/_index"
     }
 
     public func serialise() throws {
@@ -100,31 +102,30 @@ public class CreateJsonQueryIndexOperation: CouchDatabaseOperation, MangoOperati
             var jsonDict: [String:AnyObject] = ["type": "json"]
         #endif
 
-        if let fields = fields {
-            var index: [String: AnyObject] = [:]
-            #if os(Linux)
-                index["fields"] = transform(sortArray: fields).bridge()
-                jsonDict["index"] = index.bridge()
-            #else
-                index["fields"] = transform(sortArray: fields) as NSArray
-                jsonDict["index"] = index as NSDictionary
-            #endif
+        var index: [String: AnyObject] = [:]
+        #if os(Linux)
+            index["fields"] = transform(sortArray: fields).bridge()
+            jsonDict["index"] = index.bridge()
+        #else
+            index["fields"] = transform(sortArray: fields) as NSArray
+            jsonDict["index"] = index as NSDictionary
+        #endif
             
-        }
+        
 
-        if let indexName = indexName {
+        if let name = name {
             #if os(Linux)
-                jsonDict["name"] = indexName.bridge()
+                jsonDict["name"] = name.bridge()
             #else
-                jsonDict["name"] = indexName as NSString
+                jsonDict["name"] = name as NSString
             #endif
         }
 
-        if let designDoc = designDoc {
+        if let designDocumentID = designDocumentID {
             #if os(Linux)
-                jsonDict["ddoc"] = designDoc.bridge()
+                jsonDict["ddoc"] = designDocumentID.bridge()
             #else
-                jsonDict["ddoc"] = designDoc as NSString
+                jsonDict["ddoc"] = designDocumentID as NSString
             #endif
         }
 
@@ -141,6 +142,25 @@ public class CreateJsonQueryIndexOperation: CouchDatabaseOperation, MangoOperati
   A struct to represent a field in a Text index.
  */
 public struct TextIndexField {
+    
+    /**
+     The data types for a field in a Text index.
+     */
+    public enum `Type` : String {
+        /**
+         A Boolean data type.
+         */
+        case boolean = "boolean"
+        /**
+         A String data type.
+         */
+        case string = "string"
+        /**
+         A Number data type.
+         */
+        case number =  "number"
+    }
+    
     /**
      The name of the field
     */
@@ -148,46 +168,28 @@ public struct TextIndexField {
     /**
      The type of field.
      */
-    public let type: TextIndexFieldType
+    public let type: Type
   
-    public init(name: String, type: TextIndexFieldType) {
+    public init(name: String, type: Type) {
         self.name = name
         self.type = type
     }
 }
 
-/**
-    The data types for a field in a Text index.
- */
-public enum TextIndexFieldType : String {
-    /**
-     A Boolean data type.
-    */
-    case Boolean = "boolean"
-    /**
-     A String data type.
-    */
-    case String = "string"
-    /**
-     A Number data type.
-    */
-    case Number =  "number"
-}
+
 
 /**
  An Operation to create a Text Query (Mango) Index.
  
  Usage Example:
  ```
- let index = CreateTextQueryIndexOperation()
- index.indexName = "example"
- index.fields = [TextIndexField(name:"food", type: .String)
- index.defaultFieldAnalyzer = "english"
- index.defaultFieldEnabled = true
- index.selector = ["type": "food"]
- index.designDoc = "examples"
- index.databaseName = "exampledb"
- index.completionHandler = { (response, httpInfo, error) in 
+ let index = CreateTextQueryIndexOperation(databaseName: "exampledb",
+            name: "example",
+            fields: [TextIndexField(name:"food", type: .string),
+            defaultFieldAnalyzer: "english",
+            defaultFieldEnabled:  true,
+            selector:["type": "food"],
+            designDocumentID: "examples"){ (response, httpInfo, error) in
     if let error = error {
         // handle the error
     } else {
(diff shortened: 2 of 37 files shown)


Source/CreateQueryIndexOperation.swift
```diff
@@ -22,11 +22,9 @@ import Foundation
  
     Usage example:
     ```
-    let index = CreateJsonQueryIndexOperation()
+    let index = CreateJSONQueryIndexOperation(databaseName: "exampledb", fields: [Sort(field:"food", sort: .Desc)])
```
ricellis: this should be `.desc`

Source/CreateDatabaseOperation.swift
```diff
@@ -41,12 +46,8 @@ public class CreateDatabaseOperation: CouchOperation, JsonOperation {
     public var endpoint: String {
         get {
             // Safe to force unwrap, validation would fail if this is nil
```
ricellis: This comment is no longer valid.

rhyshort: removed in 72a80a2

Source/CreateQueryIndexOperation.swift
rhyshort: fixed in 72a80a2

Source/CreateDatabaseOperation.swift
```diff
@@ -19,18 +19,23 @@ import Foundation
...
     
     public var completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: ErrorProtocol?) -> Void)?
     
     /**
      The name of the database to create.
-
-     This is required to be set before the operation can execute succesfully.
      */
-    public var databaseName: String?
+    public var name: String
```
ricellis: I think the required values that are assigned during initialization should be constants. Do they need to be public?

Source/CreateQueryIndexOperation.swift
```diff
@@ -39,11 +37,19 @@ import Foundation
...
-public class CreateJsonQueryIndexOperation: CouchDatabaseOperation, MangoOperation, JsonOperation {
+public class CreateJSONQueryIndexOperation: CouchDatabaseOperation, MangoOperation, JSONOperation {
     
-    public init() { }
+    /**
+     Creates the operation
+     - parameter databaseName : The name of the database where the index should be created.
+     - parameter fields : the fields to be indexed.
+    */
+    public init(databaseName: String, fields: [Sort]) {
```
ricellis: It seems kind of strange now that the required fields are in the init, but optional fields are in properties. This feels like a bit of a programming context switch for the user.
```swift
let op = CreateJSONQueryIndexOperation(databaseName: "mydb", fields: Sort(field: "food", sort: .desc))
op.indexName = "myIndex"
op.designDoc = "myDdoc"
```

Perhaps we should consider using default values and moving all the options into the init.

Source/CreateQueryIndexOperation.swift
```diff
@@ -39,11 +37,19 @@ import Foundation
...
+        self.databaseName = databaseName
+        self.fields = fields
+    }
     
-    public var databaseName: String?
+    public var databaseName: String
```
ricellis: I guess leaving these public (even if they are constants) at least allows the values to be read if desired.

Source/CreateDatabaseOperation.swift
rhyshort: I don't see a problem with them being public, but yes making them `let` instead would make it easier for us when debugging.

Source/CreateQueryIndexOperation.swift
rhyshort: Agreed, placing everything in the init will make it more consistent. Using the defaults will also keep the list short for those who don't want all the options.

Source/CreateQueryIndexOperation.swift
```diff
@@ -141,52 +138,54 @@ public class CreateJsonQueryIndexOperation: CouchDatabaseOperation, MangoOperati
...
 public struct TextIndexField {
+    
+    /**
+     The data types for a field in a Text index.
+     */
+    public enum `Type` : String {
+        /**
+         A Boolean data type.
+         */
+        case Boolean = "boolean"
```
ricellis: These enum cases should be lowerCamelCase.

Source/CreateQueryIndexOperation.swift
```diff
@@ -141,52 +138,54 @@ public class CreateJsonQueryIndexOperation: CouchDatabaseOperation, MangoOperati
...
 
 /**
  An Operation to create a Text Query (Mango) Index.
  
  Usage Example:
  ```
- let index = CreateTextQueryIndexOperation()
+ let index = CreateTextQueryIndexOperation(databaseName: "exampledb")
  index.indexName = "example"
  index.fields = [TextIndexField(name:"food", type: .String)
```
ricellis: Will need updating to `.string`

Source/DeleteAttachmentOperation.swift
```diff
@@ -24,11 +24,10 @@ import Foundation
  
  Example usage:
  ```
- let deleteAttachment = DeleteAttachmentOperation()
- deleteAttachment.docId = docId
- deleteAttachment.revId = revId
- deleteAttachment.attachmentName = "myAwesomeAttachment"
- deleteAttachment.databaseName = "exampledb"
+ let deleteAttachment = DeleteAttachmentOperation(name: "myAwesomeAttachment",
```
ricellis: The guidelines do say to avoid abbreviations and I know you proposed `documentID` at the beginning of #96. I think that would be ok, whilst "doc" is a well known abbreviation I think since we use the term "Document" in operations, for example, there is value in being consistent. Although it is more to type the IDE will auto-complete at least.
I also think we should probably go with `revision` rather than `revID` - the Couch documentation nearly always refers to it as a "revision" (occasionally as a "revision token" and rarely as a "revision ID"), some other sources talk of a "revision number". WDYT?

Source/GetAllDocsOperation.swift
```diff
@@ -126,14 +124,14 @@ public class GetAllDocsOperation : CouchOperation, ViewOperation, JsonOperation
         }
         
         if let key = key {
-            keyJson = try convertJson(key: key)
+            keyJson = try jsonValue(for: key)
```
ricellis: In #96 you proposed `jsonValue(of: key)` was there some reasoning behind the switch to `for`. I prefer `of` or maybe `from` as `for` sounds very iterative.

Source/DeleteAttachmentOperation.swift
ricellis: I noticed later that PutAttachment is already using `documentID` so this definitely should change to match.

Source/PutAttachmentOperation.swift
```diff
@@ -25,13 +25,12 @@ import Foundation
...
- putAttachment.docId = docId
- putAttachment.revId = revId
- putAttachment.data = attachment.data(using: NSUTF8StringEncoding, allowLossyConversion: false)
- putAttachment.attachmentName = "myAwesomeAttachment"
- putAttachment.contentType = "text/plain"
- putAttachment.databaseName = "exampledb"
+ let putAttachment = PutAttachmentOperation(name: "myAwsomeAttachment",
+                                     contentType: "text/plain",
+                                            data: attachment.data(using: .utf8, allowLossyConversion: false)
+                                      documentID: docId
```
ricellis: this example var name should probably be docID to match what has been used for revID and the expected style.

Source/QueryViewOperation.swift
```diff
@@ -50,26 +48,27 @@ import Foundation
...
+    
+    public var databaseName: String
 
     /**
      The name of the design document which contains the view.
-
-     - Note: must be set for a query view operation to successfully run.
      */
-    public var designDoc: String? = nil
+    public var designDocument: String
```
ricellis: I wonder if some additional description here would make this more usable for newcomers:
```swift
QueryView(named viewName: String, fromDesignDocument designDocumentID: String, withinDatabase databaseName: String)`
let queryViewOp = QueryView(named: "myview", fromDesignDocument: "myDesignDocument", withinDatabase: "myDatabase")
```
or
```swift
QueryView(named viewName: String, from designDocumentID: String, within databaseName: String)`
let queryViewOp = QueryView(named: "myview", from: "myDesignDocument", within: "myDatabase")
```

Source/GetAllDocsOperation.swift
rhyshort: Yeah, it didn't read quite right when I started using it at the call sites. I thought that "JSON value for key" worked better than "JSON value of key". I'm not quite so sure about "JSON value from key", as that would imply its getting the JSON value from the key object rather than making a new value, but that's just me :)

Source/CreateQueryIndexOperation.swift
rhyshort: done in 0808851

Source/CreateQueryIndexOperation.swift
rhyshort: done in 0808851

Source/GetAllDocsOperation.swift
ricellis: ok after the straw poll it's clear I'm in the minority, leave it as `for`.

Source/DeleteAttachmentOperation.swift
rhyshort: Yeah I prefer `revision` to `revisionID`, `documentID` is probably the way to go, but if we are using CouchDB docs as the defining material, it refers to `documentID` as `docid` so `docID` would be more in keeping with that approach.

rhyshort: changed to `documentID` in bfa3f2c

ricellis:
A few things outstanding from #96 and review:

1. `revID` to `revision`
1. `Create*QueryIndexOperation`
    * `indexName` to `name`
    * `designDoc` to `designDocumentID`
1. `GetDocumentOperation`
    * `revs` to `includeRevisions`
1. `QueryViewOperation` & `ViewOperation`
    * `updateSeq`to `includeLastUpdateSequenceNumber`
1. `HTTPInterceptor`
    * `ctx:` to `in context:`

I'm not sure how much of that you wanted to defer to a later PR, but I think point 1 probably makes sense to do here given the extent of the doc/comment update changes for documentID anyway.

ricellis:
and the CouchDBOperation changes, but you explicitly called them out as not part of this PR.

rhyshort:
@ricellis Done all the ones you pointed out, and `CouchDBOperations`? I remember about the `CouchClient.add(operation:)` methods not going in here.

Source/SessionCookieInterceptor.swift
```diff
@@ -75,7 +75,7 @@ public class SessionCookieInterceptor: HTTPInterceptor
         return ctx;
     }
 
-    public func interceptRequest(ctx: HTTPInterceptorContext) -> HTTPInterceptorContext {
+    public func interceptRequest(in ctx: HTTPInterceptorContext) -> HTTPInterceptorContext {
```
ricellis: missed one

Source/CreateQueryIndexOperation.swift
```diff
@@ -42,10 +42,13 @@ public class CreateJSONQueryIndexOperation: CouchDatabaseOperation, MangoOperati
     /**
      Creates the operation
      - parameter databaseName : The name of the database where the index should be created.
+     - parameter designDocumentID : The ID of the design document where the index should be saved, 
+     if set to `nil` the server will create a new design document with a generated ID.
```
ricellis: do you think we should also say that the default value is `nil`?

rhyshort: I'm don't think so, the declaration when viewed (option+click on symbol in Xcode) shows the defaults for other frameworks like XCTest, so it should do the same for us. Although I've yet to see anything other than `x = default` but that could be because i'm running the beta.

ricellis: Yeah that is what I was wondering whether it was going to show up nicely automatically.

rhyshort: I think we should take the chance that'll show up when Xcode 8 and Swift 3 are released. If it still doesn't show the default, we can go back over and add the default in all the documentation entries.

Source/FindDocumentsOperation.swift
```diff
@@ -130,6 +130,37 @@ internal extension MangoOperation {
...
 public class FindDocumentsOperation: CouchDatabaseOperation, MangoOperation, JSONOperation {
     
+    /**
+    
+     Creates the operation.
+     
+     - parameter selector: the selector to use to select documents in the database.
+     - parameter databaseName: the name of the database the find operation should be performed on.
+     - parameter fields: the fields of a matching document which should be returned in the repsonse.
+     - parameter limt: the maximium number of documents to be returned.
```
ricellis: lim**i**t

Source/FindDocumentsOperation.swift
```diff
@@ -130,6 +130,37 @@ internal extension MangoOperation {
  */
 public class FindDocumentsOperation: CouchDatabaseOperation, MangoOperation, JSONOperation {
     
+    /**
```
ricellis: The example in lines 112-119 needs updating.

Source/GetAllDatabasesOperation.swift
```diff
@@ -36,6 +35,13 @@ import Foundation
  */
 public class GetAllDatabasesOperation : CouchOperation, JSONOperation {
     
+    /**
+     
+     Creates the operation
+     
+     - parameter databaseHandler: a handler to call for each database returned from the server.
```
ricellis: maybe we could say "optional handler to call" for this (it might be a bit more concise than "if nil ..."

rhyshort: done in 04debfa

Source/FindDocumentsOperation.swift
rhyshort: done in 04debfa

Source/FindDocumentsOperation.swift
rhyshort: done in 04debfa

Source/GetAllDocsOperation.swift
```diff
@@ -70,6 +69,33 @@ public class GetAllDocsOperation : CouchOperation, ViewOperation, JSONOperation
...
     public let includeLastUpdateSequenceNumber: Bool?
     
+    /**
+     Creates the operation
+     
+     - parameter databaseName: The name of the database from which to get all the documents.
+     - parameter descending: Should the documents be sorted in descending order.
+     - parameter endKey: When this document ID is hit stop returning results.
+     - parameter includeDocs: Include document content in the response.
+     - parameter conflucts: Include infomration about conflicted revisions.
```
ricellis: confl**i**cts

Source/GetAllDocsOperation.swift
```diff
@@ -70,6 +69,33 @@ public class GetAllDocsOperation : CouchOperation, ViewOperation, JSONOperation
...
+     - parameter key: Return the document with the specified ID.
+     - parameter keys: Return the documents with the soecified IDs.
+     - parameter limit: The number of documents the repsonse should be limited to.
+     - parameter skip: the number of documents that should be skipped before returning results.
+     - parameter startKeyDocID: the document ID from which to start the response.
+     - parameter endKeyDocId: the document ID at which to to stop retuning results.
```
ricellis: do we want to call these KeyDocumentID to conform with the others documentIDs?
In any case the casing is wrong on the end key Id -> ID

Source/GetAllDocsOperation.swift
```diff
@@ -70,6 +69,33 @@ public class GetAllDocsOperation : CouchOperation, ViewOperation, JSONOperation
...
+     - parameter includeLastUpdateSequenceNumber: Include the sequence number at which the view was last updated.
+     - parameter inclusiveEnd: Detmerines if the `endKey` and `endKeyDocId` should be included when returning documents.
+     - parameter stale: Whether stale views are ok, or should be updated after the response is returned.
+     - parameter startKey: the document ID from where to start returning documents.
+     - parameter startKeyDocId: the document ID from where to start returning documents.
```
ricellis: this is a duplicate of line 84

Source/GetAllDocsOperation.swift
```diff
@@ -70,6 +69,33 @@ public class GetAllDocsOperation : CouchOperation, ViewOperation, JSONOperation
...
+     - parameter key: Return the document with the specified ID.
+     - parameter keys: Return the documents with the soecified IDs.
+     - parameter limit: The number of documents the repsonse should be limited to.
+     - parameter skip: the number of documents that should be skipped before returning results.
+     - parameter startKeyDocID: the document ID from which to start the response.
+     - parameter endKeyDocId: the document ID at which to to stop retuning results.
+     - parameter includeLastUpdateSequenceNumber: Include the sequence number at which the view was last updated.
+     - parameter inclusiveEnd: Detmerines if the `endKey` and `endKeyDocId` should be included when returning documents.
+     - parameter stale: Whether stale views are ok, or should be updated after the response is returned.
+     - parameter startKey: the document ID from where to start returning documents.
```
ricellis: I don't see endKey in the list. In the case of all docs startKey and endKey are document IDs anyway, so I wonder whether it is worth omitting the duplication to reduce the size of the constructor for the all docs case (I know it is valid to supply both as it is a view type request, but I just can't see a use case for it since documents can't exist with the same ID).

Source/GetDocumentOperation.swift
```diff
@@ -16,8 +16,34 @@
 
 import Foundation
 
+/**
+ 
+ Gets a document from the server.
```
ricellis: server is an awkward word here, "Gets a document from a database"?

Source/GetDocumentOperation.swift
```diff
@@ -16,8 +16,34 @@
...
+ }
+ ```
+ 
+ */
 public class GetDocumentOperation: CouchDatabaseOperation, JSONOperation {
 
+    /**
+     Creates the operation.
+     
+      - parameter id: the id of the document to get from the server.
```
ricellis: again I don't like the use of "server", especially with db next and multiple nodes and so on, what really counts as the server?

Source/GetAllDocsOperation.swift
rhyshort: I think we change the capitalisation, I'm not keen on `startKeyDocumentID` but I can't think of a better alternative that gets the same point across.

Source/PutDocumentOperation.swift
```diff
@@ -16,8 +16,33 @@
...
+ */
 public class PutDocumentOperation: CouchDatabaseOperation, JSONOperation {
     
+    /**
+     Creates the operation.
+     
+     - parameter id: the id of the document to create or update
+     - parameter revison: the revison of the document to update.
+     - parameter body: the body of the document
+     - parameter databaseName: the name of the database where the document should/is saved.
```
ricellis: this needs to be "should be/is" but it is a bit awkward. maybe:
"the name of the database where the document will be created or updated"

Source/PutDocumentOperation.swift
```diff
@@ -37,7 +62,8 @@ public class PutDocumentOperation: CouchDatabaseOperation, JSONOperation {
     public let id: String
 
     /**
-     If updating a document, set this value to the current revision ID.HTTPInfo
+     If the operation is updating a document, this will be set to the revision of the document that
```
ricellis: I don't think "will be" is the right tense here. The property can't be accessed until after init at which point it will either have the value or it won't.
"this **is** the revision of the document that is being updated"?

Source/GetAllDocsOperation.swift
```diff
@@ -70,6 +69,33 @@ public class GetAllDocsOperation : CouchOperation, ViewOperation, JSONOperation
...
+     - parameter startKeyDocId: the document ID from where to start returning documents.
+     - parameter rowHandler: a handler to call for each row returned from the view.
+     - parameter completionHandler: a handler to call when the operation completes.
+     
+     - warning: `stale` is an advanced option, it should not be used unless you fully understand the outcome of changing the value of this property.
+     - warning: The option `key` and `keys` cannot be used together.
+     - warning: The `group` option is only valid when a `reduce` function is used.
```
ricellis: These group options aren't valid for _all_docs

Source/GetAllDocsOperation.swift
rhyshort: done in 68833ce

Source/GetAllDocsOperation.swift
rhyshort: fixed in 68833ce

Source/QueryViewOperation.swift
```diff
@@ -50,8 +50,45 @@ import Foundation
...
+     
+     - warning: `stale` is an advanced option, it should not be used unless you fully understand the outcome of changing the value of this property.
+     - warning: The option `key` and `keys` cannot be used together.
+     - warning: The `group` option is only valid when a `reduce` function is used.
+     - warning: The `groupLevel` option is only valid if `group` is `true`.
+     
+     */
     public init(name: String,
-                desginDocument: String,
+                desginDocumentID: String,
```
ricellis: typo, desgin -> design throughout this class

Source/ViewLikeOperation.swift
```diff
@@ -149,18 +117,17 @@ public protocol ViewOperation : CouchDatabaseOperation {
...
-     - Note: Optional, if the property is unset this parameter will be omitted from the request and the server 
-     default will apply.
      - SeeAlso: `Stale` for descriptions of the available values for allowing stale views.
-     - Warning: This is an advanced option, it should not be used unless you fully understand the outcome of 
-     changing the value of this property.
      */
     var stale: Stale? { get }
     
+    /**
+     Determins if the last su
```
ricellis: need to finish this

Source/GetAllDocsOperation.swift
rhyshort: I agree its not a great thing to have both, but with the aim of this part of swift cloudant to be as close to the HTTP API as possible, and since CouchDB lets you specify both, we should as well, even if it is useless.

Tests/SwiftCloudant/QueryViewTests.swift
```diff
@@ -51,7 +51,7 @@ public class QueryViewTests: XCTestCase {
     }
 
     func testViewGeneratesCorrectRequestUsingStartAndEndKeys() throws {
-        let view = QueryViewOperation(name: "view1", desginDocument: "ddoc", databaseName: self.dbName!,
+        let view = QueryViewOperation(name: "view1", desginDocumentID: "ddoc", databaseName: self.dbName!,
```
ricellis: this test class needs updating to match the type fix desgin -> design

Source/GetAllDocsOperation.swift
rhyshort: also endkey is on line 77.

Source/PutDocumentOperation.swift
rhyshort: fixed in 58ffbcd

Source/PutDocumentOperation.swift
rhyshort: fixed in 58ffbcd

Source/GetDocumentOperation.swift
rhyshort: fixed in 58ffbcd

Source/GetDocumentOperation.swift
rhyshort: fixed in 58ffbcd

Source/GetAllDocsOperation.swift
rhyshort: fixed in ed66b82

Source/QueryViewOperation.swift
rhyshort: fixed in e5ccba8

Source/ViewLikeOperation.swift
rhyshort: done in c1d82c6

Tests/SwiftCloudant/QueryViewTests.swift
rhyshort: done in ac22100

Source/GetAllDocsOperation.swift
rhyshort: fixed in 8b05c23

Source/GetAllDocsOperation.swift
ricellis: ah yes, so it is.
I guess my point was if you specify `startKeyDocumentID`, but not `startKey` - I'm not sure if that actually works. So it would be a bit confusing to have to supply a document ID to `startKey` when there is a parameter called `startKeyDocumentID` that wouldn't work, if specifying startKeyDocumentID alone works then it is less confusing.

ricellis: Also can we group the parameter descriptions so they are a bit more related, maybe startKey and endKey should be closer together for example

ricellis:
+1

PR #116, opened 2016-08-08
Support Swift 2016-08-04 Snapshot

### LINKED ISSUE #115: Support 2016-08-04 Snapshot
A new Snapshot has been released. It contains various changes such as the function arguments not requiring argument labels. The biggest change is that objc `id` is now imported as `Any` this means we can now change the API to take `Any` rather than `AnyObject`

rhyshort:
# What

Support Swift 2016-08-04 Snapshot.

## How

- Add `Tests` suffix to test module
- Update blocks to remove argument labels
- Use `Any` instead of AnyObject.

## Testing

All tests pass

## Issues

Fixes #115 

## Notes

The CI build fails due to cocoapods validation, I think this might be due the renaming of OS X to macOS in xcode 8.


Source/CreateQueryIndexOperation.swift
```diff
@@ -220,9 +220,9 @@ public class CreateTextQueryIndexOperation: CouchDatabaseOperation, MangoOperati
                       fields: [TextIndexField]? = nil,
         defaultFieldAnalyzer: String? = nil,
          defaultFieldEnabled: Bool? = nil,
-                    selector: [String:AnyObject]? = nil,
+                    selector: [String:Any]? = nil,
                    designDocumentID: String? = nil,
-           completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: Error?) -> Void)? = nil) {
+           completionHandler: (([String : AnyObject]?, HTTPInfo?, Error?) -> Void)? = nil) {
```
brynh: Shouldn't this be `[String : Any]?`?
Also, typo at line 214: `paerameter`

Source/CreateQueryIndexOperation.swift
```diff
@@ -234,7 +234,7 @@ public class CreateTextQueryIndexOperation: CouchDatabaseOperation, MangoOperati
     }
     
     public let databaseName: String
-    public let completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: Error?) -> Void)?
+    public let completionHandler: ((_ response: [String : AnyObject]?, _ httpInfo: HTTPInfo?, _ error: Error?) -> Void)?
```
brynh: Shouldn't this be `[String : Any]?`?

Source/DeleteDocumentOperation.swift
```diff
@@ -42,14 +42,14 @@ public class DeleteDocumentOperation: CouchDatabaseOperation, JSONOperation {
      - parameter databaseName: the name of the database which contains the document.
      - parameter completionHandler: optional handler to run when the operation completes.
      */
-    public init(id: String, revision: String, databaseName: String, completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: Error?) -> Void)? = nil) {
+    public init(id: String, revision: String, databaseName: String, completionHandler: (([String : AnyObject]?, HTTPInfo?, Error?) -> Void)? = nil) {
```
brynh: Shouldn't this be `[String : Any]?`?

Source/FindDocumentsOperation.swift
```diff
@@ -380,13 +380,13 @@ public class FindDocumentsOperation: CouchDatabaseOperation, MangoOperation, JSO
...
             for doc: [String: AnyObject] in docs {
-                self.documentFoundHandler?(document: doc)
+                self.documentFoundHandler?(doc)
             }
         }
     }
     
     public func callCompletionHandler(response: Any?, httpInfo: HTTPInfo?, error: Error?) {
-        self.completionHandler?(response: response as? [String: AnyObject], httpInfo: httpInfo, error: error)
+        self.completionHandler?(response as? [String: AnyObject], httpInfo, error)
```
brynh: Shouldn't this be `[String: Any]`

Source/DeleteDocumentOperation.swift
```diff
@@ -42,14 +42,14 @@ public class DeleteDocumentOperation: CouchDatabaseOperation, JSONOperation {
...
         self.id = id
         self.revision = revision
         self.databaseName = databaseName
         self.completionHandler = completionHandler
     }
     
-    public let completionHandler: ((response: [String : AnyObject]?, httpInfo: HTTPInfo?, error: Error?) -> Void)?
+    public let completionHandler: (([String : AnyObject]?, HTTPInfo?, Error?) -> Void)?
```
brynh: Shouldn't this be `[String : Any]?`?

brynh:
I don't really see a benefit in removing the argument labels.  I like argument labels.  To me, it seems like sacrificing clarity. Is there some benefit in the Swift world that I'm missing?

rhyshort:
Swift proposal [SE0111](https://github.com/apple/swift-evolution/blob/master/proposals/0111-remove-arg-label-type-significance.md) removed argument labels from the type system for functions (I think this is what removed the labels from closures too). I think its a shame that labels have been removed, we could keep them in the definition, but they don't show up in autocomplete, documentation or at the call site. I'm at a bit of a loss as what provides the best clarity, none and have them documented accordingly, or keep the labels in the definition, but a user never sees them and nor do we when using them.

Source/CreateQueryIndexOperation.swift
rhyshort: Done in 524aee4 (typo fix coming later)

Source/CreateQueryIndexOperation.swift
rhyshort: Done in 524aee4

Source/DeleteDocumentOperation.swift
rhyshort: Done in 524aee4

Source/FindDocumentsOperation.swift
rhyshort: Done in 524aee4

Source/DeleteDocumentOperation.swift
rhyshort: Done in 524aee4

Source/CreateQueryIndexOperation.swift
rhyshort: typo done in 9a229b5

brynh:
Thanks for the link to SE0111 - that's useful info.  It seems that things have changed fairly frequently around how labels work in Swift. It's probably best to leave it as it is for the moment. I think in these cases, labels don't add as much as they would for some argument lists anyway.

brynh:
+1

PR #118, opened 2016-08-22
Add the ability to page through views

### LINKED ISSUE #62: View paging
Whilst it is possible to do paging at the moment by using multiple `QueryViewOperation`s and adjusting the parameters on each operation it would be nice to codify the view pagination recipe.

One possibility would be to provide a `PagedQueryViewOperation` (name STC) that had functions that can return new `PagedQueryViewOperation` instances that can be added to the database operation queue to get the next/previous pages.

rhyshort:
## What

Added support to page through views.

## How

Created a new class `ViewPage` which handles the requests to correctly page, this can either be one page after another using the block, or later using the token.

## Testing

Added new tests to ViewPageTests.

## Issues

Resolves #62

## Notes
There isn't any API doc yet, I want to get some feedback on the API etc before documenting it.

### DIFF
--- a/Source/ViewPaging.swift
+++ b/Source/ViewPaging.swift
@@ -1,0 +1,676 @@
+//
+//  ViewPaging.swift
+//  SwiftCloudant
+//
+//  Created by Rhys Short on 01/08/2016.
+//
+//  Copyright (C) 2016 IBM Corp.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+
+import Foundation
+
+
+/** 
+ ViewPager makes it possible to paginate results from a CouchDB view.
+ There are some issues to be aware of when paginating  views.
+ 
+ Views do not support pagination natively, `ViewPager` paginates views using the `startKey`,
+ `startKey_docid`, `endKey`, `endKey_docid` and `limit` parameters to return a set number
+ of results or less. Each time a page is requested using the pageHandler closure, by returning
+ `ViewPager.Page.next` or `.previous` a new request is made to the server.
+ 
+ - warning: Paginating through views using multiple requests can cause results to be missed
+ if they are inserted at the boundry of a page between page requests. 
+ Additionally multiple requests may present inconsistent results for the same apparent page when 
+ a repeat request is made (e.g. paging back to a previous page) if results have been inserted or
+ removed inbetween the page requests.
+ 
+ Usage Example:
+ 
+ ```
+ // Make the first request with a view pager.
+ let pager = ViewPager(name: "exampleView", designDocumentID: "example", database: "exampleDB, client: client) { (response, token, error) -> .Page? 
+ 
+    //process the results
+    if someCondition {
+       return nil // return nil to stop page
+    } else {
+        return .next
+    }
+ 
+ }
+ 
+ pager.makeRequest()
+ 
+ 
+ // Get the next page from a token.
+ 
+ ViewPager.next(token: token) { (response, token, error) -> .Page? in 
+ 
+    //process the results
+    if someCondition {
+        return nil // return nil to stop requesting pages
+    } else {
+        return .next // to request the next page
+    }
+ }
+ 
+ // Get the previous page from a string token
+ let strToken = try token.serialise
+ 
+ try ViewPager.next(token: strToken) { (response, token, error) -> .Page? in
+ 
+    //process the results
+    if someCondition {
+        return nil // return nil to stop page
+    } else {
+        return .previous
+    }
+ }
+ ```
+ 
+ - seealso:
+ [CouchDB Pagination Recipe](http://docs.couchdb.org/en/stable/couchapp/views/pagination.html)
+*/
+public class ViewPager {
+    /**
+     A token which contains all the information required to continue paging for a previous
+     `ViewPager`.
+     */
+    public struct Token {
+        
+        fileprivate let state: ViewPager.State
+        
+        fileprivate let descending: Bool?
+        fileprivate let startKey: Any?
+        fileprivate let startKeyDocumentID:String?
+        fileprivate let endKey: Any?
+        fileprivate let endKeyDocumentID: String?
+        fileprivate let inclusiveEnd:Bool?
+        fileprivate let key:Any?
+        fileprivate let keys:[Any]?
+        fileprivate let includeDocs:Bool?
+        fileprivate let conflicts:Bool?
+        fileprivate let stale:Stale?
+        fileprivate let includeLastUpdateSequenceNumber: Bool?
+        
+        fileprivate let name: String
+        fileprivate let designDocumentID: String
+        fileprivate let databaseName:String
+        fileprivate let pageSize:UInt
+        
+        fileprivate let client: CouchDBClient
+        
+        
+        /**
+         Creates a serialised version of this token.
+         
+         - throws: An error if serialisation fails.
+         
+         - returns: A serialised version of this token.
+        */
+        public func serialised() throws -> String {
+            
+            var dict:[String: Any] = ["name": name,
+                                        "ddoc": designDocumentID,
+                                        "db": databaseName,
+                                        "page_size": pageSize]
+            
+            
+            if let descending = descending {
+                dict["descending"] = descending
+            }
+            if let startKey = startKey {
+                dict["startkey"] = startKey
+            }
+            
+            if let startKeyDocumentID = startKeyDocumentID {
+                dict["startkey_docid"] = startKeyDocumentID
+            }
+            
+            if let endKey = endKey {
+                 dict["endkey"] = endKey
+            }
+            
+            if let endKeyDocumentID = endKeyDocumentID {
+                dict["endkey_docid"] = endKeyDocumentID
+            }
+            
+            if let inclusiveEnd = inclusiveEnd {
+                dict["inclusive_end"] = inclusiveEnd
+            }
+            
+            if let key = key {
+                dict["key"] = key
+            }
+            
+            if let keys = keys {
+                dict["keys"] = keys
+            }
+            
+            if let includeDocs = includeDocs {
+                dict["include_docs" ] = includeDocs
+            }
+            
+            if let conflicts = conflicts {
+                dict["conflicts"] = conflicts
+            }
+            
+            if let stale = stale {
+                dict["stale" ] = "\(stale)"
+            }
+            
+            if let includeLastUpdateSequenceNumber = includeLastUpdateSequenceNumber {
+                dict["update_seq"] = includeLastUpdateSequenceNumber
+            }
+            
+            dict["state"] = state.dictionary
+
+            
+            let data = try JSONSerialization.data(withJSONObject: dict)
+            return data.base64EncodedString()
+        }
+        
+        fileprivate static func from(_ string: String, with client: CouchDBClient) throws -> Token {
+            guard let data = Data(base64Encoded: string),
+            let dict = try JSONSerialization.jsonObject(with: data) as? [String: Any]
+                else {
+                    throw Error.deseralisationFailure
+            }
+            
+            let stateDict = dict["state"] as! [String: Any]
+            let state = State.from(stateDict)
+            
+            let stale: Stale?
+            if let staleString = dict["stale"] as? String {
+                switch staleString {
+                case "ok":
+                    stale = .ok
+                    break
+                case "update_after":
+                    stale = .updateAfter
+                    break
+                default:
+                    throw Error.deseralisationFailure
+                }
+            } else {
+                stale = nil
+            }
+            
+            return Token(state: state,
+                         descending: dict["descending"] as? Bool,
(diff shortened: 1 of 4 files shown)


Source/ViewPaging.swift
```diff
@@ -0,0 +1,532 @@
+//
+//  ViewPaging.swift
+//  SwiftCloudant
+//
+//  Created by Rhys Short on 01/08/2016.
+//
+//  Copyright (c) 2016 IBM Corp.
```
ricellis: (C) or the symbol - you really need to update your template :)

Source/ViewPaging.swift
```diff
@@ -0,0 +1,532 @@
...
+            
+            if let includeLastUpdateSequenceNumber = includeLastUpdateSequenceNumber {
+                dict["update_seq"] = includeLastUpdateSequenceNumber
+            }
+            
+            dict["state"] = state.dictionary
+
+            
+            let data = try JSONSerialization.data(withJSONObject: dict)
+            if let token =  String(data: data, encoding: .utf8) {
```
ricellis: We've tended to Base64 encode tokens elsewhere, any reason not to here?

rhyshort: I didn't have a reason not to encode, and I didn't know we did that in other places :)

Source/ViewPaging.swift
rhyshort: Xcode default, not really changeable :(

rhyshort: corrected in bb40c13

Source/ViewPaging.swift
rhyshort: Done, from 3d2eda7 token strings are now Base64 encoded.

Source/ViewPaging.swift
```diff
@@ -0,0 +1,528 @@
...
+            let state = State.from(stateDict)
+            
+            let stale: Stale?
+            if let staleString = dict["stale"] as? String {
+                switch staleString {
+                case "ok":
+                    stale = .ok
+                    break
+                default:
+                    stale = .updateAfter
```
ricellis: This is a bit confusing because IIRC `stale` is a tri-state and the default behaviour is to update before returning a result, `ok` allows stale results and `updateAfter` does the update immediately after returning the stale result.
I understand that in this case it is deserializing the token so the stale parameter should be absent for the default case, but I think having "ok" and `default` as the two switches makes this harder to think about than it needs to be. Can we just declare both strings?

Source/ViewPaging.swift
```diff
@@ -0,0 +1,528 @@
...
+                conflicts:Bool? = nil,
+                stale:Stale? = nil,
+                includeLastUpdateSequenceNumber: Bool? = nil,
+                rowHandler:(([String: Any]) -> Void)? = nil,
+                pageHandler: @escaping ([String : Any]?,Token?,Swift.Error?) -> Paging) {
+        
+        self.name = name
+        self.designDocumentID = designDocumentID
+        self.databaseName = databaseName
+        self.pageSize = pageSize
```
ricellis: Because we need to be able to add 1 to this value to get a page, we should check that the supplied value doesn't exceed `max int - 1`. I know generally we allow the server to do validation, but in this particular case the server has no concept of paging and everything is a client side responsibility. If the `pageSize + 1` will result in an error rather than any kind of invisible overflow then I think it is probably ok, although would benefit from documentation of the max size.

Come to think of it all these parameters could use some documentation.

Tests/SwiftCloudantTests/ViewPagingTests.swift
```diff
@@ -0,0 +1,649 @@
...
+    }
+    
+    func testPageForward(){
+        
+        let expectation = self.expectation(description: "Paging views")
+        
+        var firstPage: [String:Any] = [:]
+        var isFirst: Bool = true
+        
+        let viewPage = ViewPage(name: "paging", designDocumentID: "paging", databaseName: dbName, client: client!, pageSize: 5){ (page: [String : Any]?, token: ViewPage.Token?, error: Error?) -> ViewPage.Paging in
```
ricellis: `ViewPage.Paging` seems a really confusing name in this context. It kind of feels like it should be a true/false on whether it is paging or not.
I wish I could think of something better.

Source/ViewPaging.swift
rhyshort: We can, but what happens in the event that we have a stale String in the token that isn't `ok` or `updateAfter`?

Source/ViewPaging.swift
rhyshort: From a quick experiment with swift 3 repl, in the event that `Int.max` is used for page size when we attempt to add 1 and stuff it into an Int, it'll fail.

```
  1> let a:Int = Int.max
a: Int = 9223372036854775807
  2> let b: Int = a + 1
b: Int = <extracting data from value failed>

Execution interrupted. Enter code to recover and continue.
Enter LLDB commands to investigate (type :help for assistance.)
```

Source/ViewPaging.swift
ricellis:
```swift
default:
   stale = nil
```
?

Source/ViewPaging.swift
ricellis: Wow that is a spectacularly unhelpful error message. I guess as long as we've documented what is allowed then even that uninformative message could be ok.

Source/ViewPaging.swift
rhyshort: I think an error would be better, since stale is mangled, we can't exactly trust any of the token. Any suggestions on a name?

rhyshort: done in 95068f0

Source/ViewPaging.swift
rhyshort: I'll doc it when the API/ general implementation seem ok

Tests/SwiftCloudantTests/ViewPagingTests.swift
rhyshort: What if we Rename the class to `View` and then the enum to `View.Page` but that might bring its own assumptions, we could also work off the assumption that if page size is zero the whole view is returned. Thoughts?

rhyshort: Thinking about this a big more, `ViewPager` is more descriptive about what the class is for. Then we have `ViewPager.Paging` which is still a bit off, but I'm still at a loss as to wait to call it, `Action` seems a bit too vague for this 😕 . Perhaps `ViewPager.Page` and then we transform `.stop` into `.none`. Thoughts @ricellis

ricellis:
What about just using `nil` for the termination of the paging? 
So you can either to `.next` `.previous` or `nil` which stops us trying to make the enum mean different things. Otherwise I think `ViewPager.Action` `getNextPage` `getPreviousPage` `stop` might work better as those are the instructions that get returned.

rhyshort:
@ricellis `nil` now terminates paging, documentation has been added too.

Source/ViewPaging.swift
```diff
@@ -19,8 +19,76 @@
 import Foundation
 
 
+/** 
+ ViewPager makes it possible to paginate results from a CouchDB view.
+ There are some issues to be aware of when paginating  views. Views do not support
+ pagination natively, there are a two main techniques for view pagination:
+ 
+ 1. Retrieve the entire view result and paginate in memory.
+ 1. Use multiple requests with `startKey`, `startKey_docid`, `endKey`, `endKey_docid` and `limit`
```
ricellis: I would swap the order of these two points so that the one being used is first.

rhyshort: done in 726955f

Source/ViewPaging.swift
```diff
@@ -19,8 +19,76 @@
...
+ parameters. (This is the option `ViewPager` uses).
+ 
+ 
+ - warning: Paginating through views using multiple requests means results which match the 
+ view parameters if they are inserted into the database **after** the page on which they would appear
+ has been requested.
+ 
+ `ViewPager` paginates views using the `startKey`, `startKey_docid`, `endKey`, `endKey_docid` and `limit`
+ parameters, this allows the amount of memory used by paging through views to be reduced, however it
+ has the downside of not being able access a specified page, or provide a page count calculation.
```
ricellis: I wonder if we could do away with this paragraph which duplicates some of what is said in the 1/2 bullets above. I propose adding to the "memory" bullet something along the lines of:
"Use this approach if you want to be able to jump to a specific page in the result set or calculate the total number of pages. Note that depending on the number of results and content a large amount of memory may be needed to store a complete result set."
WDYT?

Source/ViewPaging.swift
```diff
@@ -19,8 +19,76 @@
...
+ 
+ Usage Example:
+ 
+ ```
+ // Make the first request with a view pager.
+ let pager = ViewPager(name: "exampleView", designDocumentID: "example", database: "exampleDB, client: client) { (response, token, error) -> .Page? 
+ 
+    //process the results
+    if someCondition {
+       return nil // return nil to stop page
```
ricellis: "stop paging"?
or maybe "to stop requesting pages"?

Source/ViewPaging.swift
```diff
@@ -19,8 +19,76 @@
...
+    } else {
+        return .next
```
ricellis: // to request the next page

Source/ViewPaging.swift
```diff
@@ -254,6 +347,35 @@ public class ViewPager {
     /// MARK: state properties for generating the next page etc.
     private var state: State = State()
     
+    
+    /**
+     Creates a view pager
+    
+     - parameter name:                            The name of the view
+     - parameter designDocumentID:                The ID of the design document which contains the view.
```
ricellis: with or without the `_design` prefix?

Source/ViewPaging.swift
```diff
@@ -357,7 +482,9 @@ public class ViewPager {
             inclusiveEnd = self.inclusiveEnd
         }
         
-        
+        // This deliberately causes a retain cycle for `self`. This is so the async class
+        // methods do not need to return the ViewPager to be retain by calling classes.
```
ricellis: retain**ed**

Source/ViewPaging.swift
rhyshort: I think we need something in the `warning` section since it will jump out more in any generated documentation, even if it does duplicate some of the other doc. Any suggestions for what we should limit it to?

Source/ViewPaging.swift
rhyshort: fixed in 8151246

Source/ViewPaging.swift
rhyshort: without `_design` prefix, fixed ie52ad21

Source/ViewPaging.swift
rhyshort: fixed in 8151246

Source/ViewPaging.swift
rhyshort: fixed in 8151246

Source/ViewPaging.swift
rhyshort: added doc note in 30da851

Tests/SwiftCloudantTests/ViewPagingTests.swift
rhyshort: Transformed to `ViewPager.Page` and return value from closure to `.Page?`

Source/ViewPaging.swift
```diff
@@ -21,21 +21,15 @@ import Foundation
...
- parameters. (This is the option `ViewPager` uses).
- 1. Retrieve the entire view result and paginate in memory.
-
- 
- - warning: Paginating through views using multiple requests means results which match the 
- view parameters if they are inserted into the database **after** the page on which they would appear
- has been requested.
+ Views do not support pagination natively, `ViewPager` paginates views using the `startKey`,
+ `startKey_docid`, `endKey`, `endKey_docid` and `limit` parameters to return a set number
+ of results or less. Each time a page is requested using the pageHandler clousre, by retirning
```
ricellis: clousre -> closure
retirning -> returning

rhyshort: fixed in 4ebd1c4

Source/ViewPaging.swift
```diff
@@ -21,21 +21,15 @@ import Foundation
...
+ `ViewPager.Page.next` or `.previous` a new request is made to the server.
  
- `ViewPager` paginates views using the `startKey`, `startKey_docid`, `endKey`, `endKey_docid` and `limit`
- parameters, this allows the amount of memory used by paging through views to be reduced, however it
- has the downside of not being able access a specified page, or provide a page count calculation.
+ - warning: Paginating through views using multiple requests can cause results to be missed
```
ricellis: I think there is another case of inconsistent results besides being missed in a boundary which is that you might expect a page you have been to before to look the same if you went back to it, but of course it could change.

What do you think about adding a bit for the case where extra results appear as well
>Additionally multiple requests may present inconsistent results for the same apparent page when a repeat request is made (e.g. paging back to a previous page) if results have been inserted or removed inbetween the page requests.

rhyshort: added in 25ba0c2

Source/ViewPaging.swift
```diff
@@ -29,7 +29,10 @@ import Foundation
  `ViewPager.Page.next` or `.previous` a new request is made to the server.
  
  - warning: Paginating through views using multiple requests can cause results to be missed
- if they are inserted at the boundry of a page between page requests.
+ if they are inserted at the boundry of a page between page requests. 
+ Additionally multiple requests may present inconsistent results for the same apparent page when
```
ricellis: needs an "a" sorry!
when **a** repeat request

ricellis:
+1 after adding that a

Source/ViewPaging.swift
rhyshort: done in 35806e9

PR #122, opened 2016-08-24
Support the Changes Feed

### LINKED ISSUE #26: Support The Changes Feed
Create an operation to support reading the changes feed (`{db}/_changes`). This will need to support the following use cases:

1. a continuous  feed, this will have a block that is trigger for each change.
2. One shot changes feed.

rhyshort:
## What

Support the Changes feed.

## How

Created new `GetChangesOperation` API. There are some limitations, for example you cannot stream the changes feed from the server. The current only supported option is a one shot request to get the latest changes, this can be combined with the `since` parameter to get on a subset of the changes in future.

## Testing

Added GetChangesTests which contains tests for checking the HTTP properties and two full requests to CouchDB.

## Issues

Resolves #26

### DIFF
--- a/Source/GetChangesOperation.swift
+++ b/Source/GetChangesOperation.swift
@@ -1,0 +1,258 @@
+//
+//  GetChangesOperation.swift
+//  SwiftCloudant
+//
+//  Created by Rhys Short on 22/08/2016.
+//
+//  Copyright (C) 2016 IBM Corp.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+
+import Foundation
+
+/**
+ Operation to get the changes feed for a database.
+ 
+ Example usage:
+ ```
+ let changes = GetChangesOperation(dbName: "exampleDB", changeHandler: {(change) in 
+    // do something for each change.
+ }) { (response, info, error) in 
+    if let error = error {
+        // handle the error
+    } else {
+        // process the changes feed result.
+    }
+ }
+ ```
+ */
+public class GetChangesOperation : CouchDatabaseOperation, JSONOperation {
+    
+    public typealias Json = [String: Any]
+
+    /**
+     Abstract representation of a CouchDB Sequence. No assumptions should be
+     made about the concrete type that is returned from the server, it should be used
+     transparently.
+     */
+    public typealias Sequence = Any
+    
+    /**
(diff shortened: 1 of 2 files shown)


Source/GetChangesOperation.swift
```diff
@@ -0,0 +1,247 @@
...
+    
+    public let completionHandler: (([String: Any]?, HTTPInfo?, Error?) -> Void)?
+    
+    /**
+     List of document IDs to limit the changes to.
+     */
+    public let docIDs: [String]?
+    
+    /**
+     Include conflict information in the response
```
emlaver: This and several other comments don't end in a period.  Looking at the rest of the library, it's best to check and add periods in this class.

Source/GetChangesOperation.swift
```diff
@@ -0,0 +1,247 @@
...
+     The view to use for filtering the changes feed, to be used with the `_view` filter.
+     */
+    public let view: String?
+    
+    /**
+     A handler to run for each entry in the `results` array in the response.
+     */
+    public let changeHandler: (([String: Any]) -> Void)? // this will be each result in the `results` section of the response
+    
+    public init(databaseName:String,
```
emlaver: There should be parameter descriptions like in the python library: https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L453

Source/GetChangesOperation.swift
```diff
@@ -0,0 +1,247 @@
...
+                return "GET"
+            }
+        }
+    }
+    
+    
+    
+    
+    
+
```
emlaver: Remove extra space.

emlaver:
What about test cases for invalid parameter types or invalid style value?

rhyshort:
Style cannot have an invalid value since its a enum the compiler will fail to compile the code. There also isn't a validation of parameters beyond that they are the right type by the compiler.

Source/GetChangesOperation.swift
rhyshort: done in b288de3

Source/GetChangesOperation.swift
rhyshort: done in b288de3

Source/GetChangesOperation.swift
emlaver: Will you add a period here?

Source/GetChangesOperation.swift
```diff
@@ -0,0 +1,258 @@
...
+     */
+    public let filter: String?
+    
+    /**
+     Include the associated document with each result in the changes feed.
+     */
+    public let includeDocs: Bool?
+    
+    /**
+     Include attachments as Base-64 encoded data in the document
```
emlaver: Add period here?

emlaver:
@rhyshort Thanks for explaining the style enum and parameter types.  There's two spots left with missing periods that I commented above, but aside from that it looks good to me +1.

Source/GetChangesOperation.swift
rhyshort: done in 	fb7ec40

Source/GetChangesOperation.swift
rhyshort: done in 	fb7ec40

PR #126, opened 2016-09-08
Merge Cookie handling into InterceptableSession 

### LINKED ISSUE #110: SessionCookie Interceptor retries on all 403 responses

- SwiftCloudant version(s) that are affected by this issue: All
- Swift version(s): N/A
- Platform(s): All
```swift
 if response.statusCode == 403 || response.statusCode == 401 {
                ctx.shouldRetry = true
                self.cookie = nil
            } else if let cookieHeader = response.allHeaderFields["Set-Cookie"] as? String {
                cookie = cookieHeader.components(separatedBy: ";").first
            }
```

So if the 403 response is an actual `FORBIDDEN` rather than `Session expired` the interceptor will force a retry anyway, and attempt to get a new cookie. This will cause excessive requests. In order to correctly identify the `FORBIDDEN` response as credentials have expired, we need to check the body of the response.

### LINKED ISSUE #124: Simplify Cookie Handling.
URLSession should handle all the functions around Cookie authentication apart from requesting the cookie. To this end, the interceptor functionality should be folded into `InterceptableSession`, and it should only make a session request when:

1. It is the first request for the session.
2. A task returned a Forbidden Error Code, with the credentials expired message.

However it should be noted that if getting a cookie fails for any reason other than a 401, it should
be retired for an amount of retires, configurable from the client.

In an long running application making enough requests, it should only be in rare situations where case 2) is invoked.

rhyshort:
## What

Merge Cookie handling into InterceptableSession 

## How

1. Remove Interceptor API from Swift Cloudant.
2. Add Cookie handling code into InterceptableSession.
    - If the request is the first one by the session and a username and password is set, a request to `_session` is made. 
   - If a 401 or a 403 with the json `{ "error" : "credentials_expired" }` then renew the cookie, and retry the request.
3. Use `HTTPCookie` and `HTTPCookieStorage` to extract the cookie and store it in the shared store.
    - URLSession by default will add cookies from the cookie store to the request, all we need to do is store the cookie in the cookie storage.

## Testing

New tests added in `InterceptableSessionTests` and ported `InterceptorTests` to test the session rather than the interceptor and move them to `InterceptableSessionTests` class.

## Issues

Resolves #124 
Resolves #110


rhyshort:
The CI failures are due to a differing swift version on my machine to the CI box. (I've got Xcode 8 GM toolchain). I'm updating the CI to handle this.

rhyshort:
Upgrading to xcode 8 GM fixed the issue.

Source/URLSession.swift
```diff
@@ -245,78 +240,218 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
...
+                        }
+                        
+                    } else {
+                        // we didn't get ok true, but got a successful response odd state, 
+                        // probably best to log it and make it so we don't request the cookie again.
+                        NSLog("Cookie request failed, response was not a JSON dictionary")
+                    }
+                    
+                } catch {
+                    // log the error from the desearlisation, this is an odd state, we should
```
ricellis: typo in deserialization

Tests/SwiftCloudantTests/InterceptableSessionTests.swift
```diff
@@ -20,13 +20,21 @@ import Foundation
...
+    
     lazy var sessionConfig = {() -> URLSessionConfiguration in
         let config = URLSessionConfiguration.default
-        config.protocolClasses = [BackOffHTTPURLProtocol.self]
+        config.protocolClasses = [BackOffHTTPURLProtocol.self,
+                                  CookieSession401.self,
+                                  CookieSession200.self,
+                                  CookieSession200Then403.self,
+                                  CookieSession200Then401.self,
+                                  CookieSession403.self
```
ricellis: I'd like to see an extra test for 200Then403 where the 403 doesn't have `credentials_expired`

Source/URLSession.swift
```diff
@@ -245,78 +240,218 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
...
                 self.taskDict[task.inProgressTask] = task
-                task.resume()
                 
                 completionHandler(.cancel)
+                
+                self.delegateQueue.asyncAfter(wallDeadline: deadline) {
+                    task.resume()
+                }
+                
+            } else if  response.statusCode == 401 || response.statusCode == 403 {
```
rhyshort: Well we don't know if credentials have expired or not until we receive the payload. So in order to be transparent, we have to delay sending of the response to the delegate, otherwise if the delegate supported streaming and delivered the parts to to an operation as they came available they would see a 403, when in fact we are going to authenticate again, retry the request and deliver a new response.

Source/URLSession.swift
```diff
@@ -245,78 +240,218 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
...
+            task.delegate.received(response: response)
+            completionHandler(.allow)
         }
-
         
     }
     
+    /**
+     Makes a synchronous request to the `_session` endpoint to start a session with the server.
+     This function will wait 600 seconds
```
rhyshort: 10 minute timeout, for one `_session` request, this is the value that was previously used by the CookieInterceptor in Objective-Cloudant and it's made its way here.

Tests/SwiftCloudantTests/InterceptableSessionTests.swift
rhyshort: We have that see `testSessionSessionHandles403NoReason`

Source/URLSession.swift
```diff
@@ -189,17 +168,21 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
         #else
             config.httpAdditionalHeaders = [("User-Agent" as NSString) as AnyHashable: InterceptableSession.userAgent()]
         #endif
+        config.httpCookieAcceptPolicy = .always
```
rhyshort: hmmm there is some contradiction over the default type, URLSessionConfiguration doc indicates its `.onlyFromMainDocumentDomain` (https://developer.apple.com/reference/foundation/nsurlsessionconfiguration/1408933-httpcookieacceptpolicy) but the documentation for the enum, indicates that .always is the default https://developer.apple.com/reference/foundation/nshttpcookieacceptpolicy.  The only reason why I set it to `.any` was to explicitly show what the value is for anyone reading the code.

Source/URLSession.swift
```diff
@@ -189,17 +168,21 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
...
-    
     private var taskDict: [Foundation.URLSessionTask: URLSessionTask] = [:]
     
     private let delegateQueue = DispatchQueue(label: "com.cloudant.swift.interceptable.session.delegate", attributes: .concurrent)
     
     private let configuration: InterceptableSessionConfiguration
     
+    private var isFirstRequest:Bool = true
+    
+    private let sessionRequestBody: Data? // This will be the body send the request to _session.
```
rhyshort: corrected in 305805a

Source/URLSession.swift
```diff
@@ -211,8 +194,17 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
      - parameter requestInterceptors: Interceptors to use with this session.
```
rhyshort: commits d6a19bb 0ac0a05 update the comment to refer to the configuration instead

Source/URLSession.swift
rhyshort: fixed in b6e689a

Source/URLSession.swift
ricellis: I would be tempted to change this to `.onlyFromMainDocumentDomain` since we shouldn't need to keep any other cookies, even if it is explicitly called for clarity and because the documentation doesn't make it clear exactly which is the default.

Source/URLSession.swift
ricellis: got it

Source/URLSession.swift
ricellis: seems pretty long, but ok

Tests/SwiftCloudantTests/InterceptableSessionTests.swift
ricellis: oh, I got confused the NoReason test in java-cloudant has a response body with:
`{"error" : "credentials_expired"}`
as opposed to
`{"error" : "credentials_expired", "reason" : "some reason"}`
i.e. it has an error of `credentials_expired` but no reason, which was a very specific error case in java, but quite distinct from the 403 response with a different error message.

Shouldn't it do a 200 then 403 though?

rhyshort: Ah I see, I thought it was a 401 forbidden from when I read the tests names :) It should receive 200, for the `_session` and then `403` for the actual request to the server, which I'm pretty sure it does.

ricellis: ok

Tests/SwiftCloudantTests/InterceptableSessionTests.swift
```diff
@@ -39,6 +39,13 @@ class InterceptableSessionTests : XCTestCase {
     
     }()
     
+    override func setUp() {
+        super.setUp()
+        // Since we shared cookie storage, we need to clear the cookies so we don't
+        // have tests leaving state behind.
+        HTTPCookieStorage.shared.removeCookies(since: Date.distantPast)
```
ricellis: The comment suggests tests won't leave state behind, but this is set up not tear down. I think doing it in setup is generally better in case an interruption or catastrophic error prevents teardown from running, but it could conceivably be done in both or the comment could be updated to make it clear that it is cleaning up any state previously left behind.

rhyshort: So, I've updated the comment and added a `teardown` because I realised that the cookies would be left behind for the next test suite, and that it may result in Authentication failures since the cookie string isn't exactly well crafted to limit it to `usernamexxx.cloudant.com`, see commit 3b754b1

Tests/SwiftCloudantTests/InterceptableSessionTests.swift
```diff
@@ -41,8 +41,13 @@ class InterceptableSessionTests : XCTestCase {
...
         super.setUp()
-        // Since we shared cookie storage, we need to clear the cookies so we don't
-        // have tests leaving state behind.
+        // Since we share cookie storage, we need to clear up any state previously left behind
+        HTTPCookieStorage.shared.removeCookies(since: Date.distantPast)
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        // Since we shared cookie storage, we need to clear up any state previously left behind
```
ricellis: or in this case try not to leave any state behind

PR #136, opened 2016-10-31
Fix issue where 401 response triggers infinite loop.

### LINKED ISSUE #135: When 401 is encountered Swift Cloudant will enter an infinite loop.

- SwiftCloudant version(s) that are affected by this issue. : 0.6.0 (haven't checked previous versions)
- Swift version(s) 3.0.1
- Platform(s) OS X.

Modifying the `CookieSession401` in `InterceptableSessionTests` to return a json structure with 
the `error` key will present the problem.

This is caused by a lack of counting down the retries when a 401 is encountered. a 401 may indicate that the session has expired or that the user cannot access the endpoint with the credentials provided. It is in the case where the suer cannot access the endpoint with the credentials provided that will cause the infinite loop.

rhyshort:
## What

Correct issue where constant 401 responses from an endpoint would cause an infinite retry loop.

## How
Decrement remainingRetries counter and check that is greater than zero before attempting to 
retry the request.

## Testing

Added a new test, and modified `CookieSession401` URLProtocol response to contain the `error` key.

## Issues

Fixes #135

### DIFF
--- a/Source/URLSession.swift
+++ b/Source/URLSession.swift
@@ -259,8 +259,10 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
                 let json = try? JSONSerialization.jsonObject(with: data)
                 if let json =  json as? [String: Any],
                     let  errorMessage = json["error"] as? String,
-                    ( errorMessage == "credentials_expired" || response.statusCode == 401 ) { // only 403 may have this message.
+                    ( errorMessage == "credentials_expired" || response.statusCode == 401 ),
+                    mTask.remainingRetries > 0 { // only 403 may have this message.
                     // we need to get a new cookie and retry the request.
+                    mTask.remainingRetries -= 1
                     self.requestCookie(url: mTask.request.url!)
                     
                     //clear the task cache
--- a/Tests/SwiftCloudantTests/InterceptableSessionTests.swift
+++ b/Tests/SwiftCloudantTests/InterceptableSessionTests.swift
@@ -217,6 +217,23 @@ class InterceptableSessionTests : XCTestCase {
         XCTAssertNotNil(cookie)
     }
     
+    func testSessionHandlesEndpointConstant401() throws {
+        let session = InterceptableSession(delegate: nil, configuration: InterceptableSessionConfiguration(shouldBackOff:true, username:"username", password:"password"))
+        session.session = URLSession(configuration: sessionConfig, delegate: session, delegateQueue: nil)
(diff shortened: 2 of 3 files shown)


Source/URLSession.swift
```diff
@@ -259,8 +259,10 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
                 let json = try? JSONSerialization.jsonObject(with: data)
                 if let json =  json as? [String: Any],
                     let  errorMessage = json["error"] as? String,
-                    ( errorMessage == "credentials_expired" || response.statusCode == 401 ) { // only 403 may have this message.
+                    ( errorMessage == "credentials_expired" || response.statusCode == 401 ),
```
emlaver: Should `"error":"unauthorized"` be added for dbnext? Sam's comment about this: https://github.com/cloudant/python-cloudant/issues/235#issuecomment-257272021

emlaver: Ok, is there a reason to keep `"credentials_expired"` if it's being removed in dbnext?

rhyshort: Until DBCore is retired from everywhere yes, otherwise customers on DBCore will lose the auto-renew function. We can remove it when DBNext is on all clusters and after we confirm that any supported versions of local don't produce `credentials_expired`.

emlaver: Ok sounds good.

Tests/SwiftCloudantTests/InterceptableSessionTests.swift
```diff
@@ -217,6 +217,21 @@ class InterceptableSessionTests : XCTestCase {
...
     
+    func testSessionHandlesEndpointConstant401() throws {
+        let session = InterceptableSession(delegate: nil, configuration: InterceptableSessionConfiguration(shouldBackOff:true, username:"username", password:"password"))
+        session.session = URLSession(configuration: sessionConfig, delegate: session, delegateQueue: nil)
+        
+        // create a context with a request which we can use
+        let url = URL(string: "http://username1.cloudant.com")!
+        let request = URLRequest(url: url)
+        
+        let expectation = self.expectation(description: "10 401 requests/responses")
```
ricellis: Can this test assert the number of retries?
I think there are 11 401 responses right? 1 for the initial request and then 10 retries?

rhyshort: Yes it can, added in 55223ee

PR #138, opened 2016-11-07
Make it possible to use the server to generate doc ids.

### LINKED ISSUE #97: Cannot use CouchDB to generate doc ids
- SwiftCloudant version(s) that are affected by this issue: All released
- Swift version(s): N/A
- Platform(s): All

``` swift
let put = PutDocumentOperation() 
put.body = ["hello":"world"] 
put.validate() // returns False instead of true,
```

Since Couch makes it possible to post a document without an id for one to be generated, we should be doing the same thing.

rhyshort:
## What

Made it possible to use the server to generate doc ids.

## How

Loosen the restriction on `PutDocumentOperation` for the `id` parameter,
to allow `nil` to use the sever to generate a doc id when the document
is saved.

## Testing

Several new tests added to `PutDocumentTests`

## Issues

Fixes #97

### DIFF
--- a/Source/SwiftCloudant/Operations/Documents/PutDocumentOperation.swift
+++ b/Source/SwiftCloudant/Operations/Documents/PutDocumentOperation.swift
@@ -37,13 +37,13 @@ public class PutDocumentOperation: CouchDatabaseOperation, JSONOperation {
     /**
      Creates the operation.
      
-     - parameter id: the id of the document to create or update
-     - parameter revison: the revison of the document to update.
+     - parameter id: the id of the document to create or update, or nil if the server should generate an ID.
+     - parameter revision: the revision of the document to update, or `nil` if it is a create.
      - parameter body: the body of the document
      - parameter databaseName: the name of the database where the document will be created / updated.
      - parameter completionHandler: optional handler to run when the operation completes.
      */
-    public init(id: String, revision: String? = nil, body: [String: Any], databaseName:String, completionHandler: (([String : Any]?, HTTPInfo?, Error?) -> Void)? = nil) {
+    public init(id: String? = nil, revision: String? = nil, body: [String: Any], databaseName:String, completionHandler: (([String : Any]?, HTTPInfo?, Error?) -> Void)? = nil) {
         self.id = id;
         self.revision = revision
         self.body = body
@@ -59,7 +59,7 @@ public class PutDocumentOperation: CouchDatabaseOperation, JSONOperation {
     /**
      The document that this operation will modify.
      */
-    public let id: String
+    public let id: String?
 
     /**
      The revision of the document being updated or `nil` if this operation is creating a document.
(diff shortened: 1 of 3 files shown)


Tests/SwiftCloudantTests/PutDocumentTests.swift
```diff
@@ -54,4 +54,75 @@ class PutDocumentTests: XCTestCase {
...
+        let json = try JSONSerialization.jsonObject(with: put.data!) as? [String:String]
+        
+        if let json = json {
+            XCTAssertEqual(["hello":"world"], json)
+        }else {
+            XCTAssertNotNil(json)
+        }
+    }
+    
+    func testOperationReuqestWithId() throws {
```
emlaver: Fix name to `testOperationRequestWithId`.

rhyshort: fixed in 0914b01

Source/SwiftCloudant/Operations/Documents/PutDocumentOperation.swift
```diff
@@ -37,13 +37,13 @@ public class PutDocumentOperation: CouchDatabaseOperation, JSONOperation {
     /**
      Creates the operation.
      
-     - parameter id: the id of the document to create or update
+     - parameter id: the id of the document to create or update, if `nil` the server will generate an ID.
```
emlaver: This sounds awkward and would be better as two separate sentences:
`parameter id: the id of the document to create or update.  If `nil`, the server will generate an ID.`

Tests/SwiftCloudantTests/PutDocumentTests.swift
```diff
@@ -54,4 +54,75 @@ class PutDocumentTests: XCTestCase {
...
+        XCTAssertEqual(put.endpoint, "/\(self.dbName!)")
+        XCTAssertEqual([:] as [String:String], put.parameters)
+        try put.serialise()
+        
+        let json = try JSONSerialization.jsonObject(with: put.data!) as? [String:String]
+        
+        if let json = json {
+            XCTAssertEqual(["hello":"world"], json)
+        }else {
+            XCTAssertNotNil(json)
```
brynh: I'd find this slightly clearer if it called `XCTFail` instead of `XCTAssertNotNil`. Similarly for the next 2 test cases.

Source/SwiftCloudant/Operations/Documents/PutDocumentOperation.swift
```diff
@@ -37,13 +37,13 @@ public class PutDocumentOperation: CouchDatabaseOperation, JSONOperation {
     /**
      Creates the operation.
      
-     - parameter id: the id of the document to create or update
+     - parameter id: the id of the document to create or update, or nil if the server should generate an ID.
      - parameter revison: the revison of the document to update.
```
brynh: Not directly related to this change, but `revision` is spelt incorrectly twice on this line and it also only mentions updates. I think it's worth mentioning here it can be `nil` for a create.

rhyshort: done in 86a49f9

Tests/SwiftCloudantTests/PutDocumentTests.swift
rhyshort: done in 86a49f9

PR #147, opened 2017-01-13
Add Jenkinsfile.

rhyshort:
## What

Create a Jenkinsfile for automated internal testing.

## How

- Create a jenkinsfile which can run Mac and Linux builds in parallel.
- Modify all dbs tests to check for the presence of the database created rather than for only
   that database to be present.

## Testing

Testing changes detailed above, no other changes

### DIFF
--- a/Tests/SwiftCloudantTests/GetAllDatabasesTests.swift
+++ b/Tests/SwiftCloudantTests/GetAllDatabasesTests.swift
@@ -39,13 +39,16 @@ public class GetAllDatabasesTest : XCTestCase {
     
     public func testListAllDbs(){
         let expectation =  self.expectation(description: "all_dbs")
+        var found = false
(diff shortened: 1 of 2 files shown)


Tests/SwiftCloudantTests/GetAllDatabasesTests.swift
```diff
@@ -58,13 +61,10 @@ public class GetAllDatabasesTest : XCTestCase {
                 let expected: [String] = [self.dbName!]
                 
                 // we need to filter the dbs responses to remove system databases, ones prefixed with _
```
smithsz: remove comment

rhyshort: removed in 716edb4

PR #153, opened 2017-04-06
Use KVO APIs for Operations

ianpartridge:
Now that Foundation on Linux support async Operations via KVO APIs, we can remove the Linux-specific code in Cloudant.


rhyshort:
Thanks for this Ian.

PR #155, opened 2017-04-21
Workaround lack of Bundle(for:) on Linux

ianpartridge:
As `Bundle(for:)` is not available on Linux, and is unlikely to be implemented in the short-term, work around this omission.

### DIFF
--- a/Source/SwiftCloudant/HTTP/URLSession.swift
+++ b/Source/SwiftCloudant/HTTP/URLSession.swift
@@ -457,18 +457,18 @@ internal class InterceptableSession: NSObject, URLSessionDelegate, URLSessionTas
         #else
             let platform = "Unknown";
         #endif
-        let frameworkBundle = Bundle(for: InterceptableSession.self)
-        var bundleDisplayName = frameworkBundle.object(forInfoDictionaryKey: "CFBundleName")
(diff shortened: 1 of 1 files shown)


ianpartridge:
@ricellis @rhyshort is this OK by you?

ricellis:
I'm not 100% happy having it always set to `Unknown` on Linux. We sometimes need the version number for diagnostics and metrics. IIRC there isn't a tool that we can use for both SPM and CocoaPods to do a version number replacement, which I think leaves the only other option as setting the version string manually in the `URLSession.swift` file as well as the `SwiftCloudant.podspec`. I think this extra step for a release would be preferable to always using `Unknown` assuming there isn't an automatic way for us to do it.

ianpartridge:
Thanks @ricellis I've updated the PR to specify the version number.

I used 0.6.0 as this is what `SwiftCloudant.podspec` currently says on the `linux-support` branch.  Before we merge this branch to `master` we'll need to make sure the version numbering on this branch is correct.

I can open a separate PR for that now if you'd like?

ricellis:
Sounds good, I'll update our release documentation to match.

ianpartridge:
Thanks - please merge this PR if you're happy 👍

PR #156, opened 2017-04-24
Specify 3.1.1 in .swift-version

ianpartridge:
Swift 3.1.1 is now released.  This is the version which adds support for async usage of `Operation` via `willChangeValue(forKey:)`, as used by this SDK.

Move the `linux-support` branch to this version.


ianpartridge:
@ricellis here's another small one. Might it be possible for me to get commit access here?

ricellis:
@ianpartridge I'll look into it.

In the meantime I think this needs some additional changes to the `.travis.yml` to get the newer Swift version.

ianpartridge:
Thanks @ricellis - I've updated `.travis.yml`.  The CI failed but it looks like a network issue to me.

ricellis:
The Linux travis build had this issue, which I think will need looking at:
```
/home/travis/build/cloudant/swift-cloudant/Source/SwiftCloudant/HTTP/URLSession.swift:171:37: error: type 'HTTPCookieStorage?' has no member 'shared'

        config.httpCookieStorage = .shared
```

The macOS failure was interestingly the `getChanges` tests, which had been causing problems on Linux, perhaps the failures in those tests are related to Swift 3.1.1 rather than Linux v Mac differences, but I think that is a separate investigation.

ianpartridge:
The Linux failure is known.  `HTTPCookieStorage` is not in Swift 3.1.1 (it'll be in the next release).

Can you try another run to see if the `getChanges` failure is consistent?

ianpartridge:
OK that passed 😄  I think this PR is good to go, then. It won't have affected the macOS build so the intermittent `getChanges` issue must have been there already.

We are still investigating the `getChanges` failure on Linux.

PR #160, opened 2017-04-26
Remove implicit bridging in tests

ianpartridge:
Although Darwin platforms have automatic bridging between Foundation types such as `NSDictionary` and Swift types like `Dictionary`, this is unfortunately not supported on Linux.

Work around this by explicitly bridging instead.  Although this does create additional instances of Foundation classes on Darwin, this is OK as it's only in the tests so won't affect performance.


ianpartridge:
Are you OK with this @ricellis ?

PR #162, opened 2017-06-30
Linux support

ianpartridge:
Hi @ricellis @rhyshort.

I'd like to start a conversation about merging the `linux-support` branch to `master`. We have done a lot of work over in Foundation and all the Cloudant tests now pass, including the `InterceptableSessionTests`, when you use the latest dev snapshots from `swift.org`.

Could you review this PR and let me know your thoughts on doing this merge?

Many thanks.


rhyshort:
I'm okay with the code, but I'm concerned about the build. We currently still only pass on macOS, Swift 3.1.1 on Linux is missing .shared for HTTPCookieStorage. With September not far off, I doubt there'll be another swift release before Swift 4 lands.  

So I think we are going to need to bump the versions up to 4, not sure if we should do that now or sit and wait until the official release.

ianpartridge:
Hi @rhyshort - yes Linux support will arrive with a Swift 4 release.  Do we want to bump `master` up to 4.0 then do a release of this SDK as soon as Swift 4 ships in September?

Source/SwiftCloudant/Operations/Query/FindDocumentsOperation.swift
```diff
@@ -344,17 +289,9 @@ public class FindDocumentsOperation: CouchDatabaseOperation, MangoOperation, JSO
         for s in sortArray {
             if let sort = s.sort {
                 let dict = [s.field: sort.rawValue]
-                #if os(Linux)
-                    transfomed.append(dict.bridge())
-                #else
-                    transfomed.append(dict )
-                #endif
+                transfomed.append(dict )
```
ricellis: The trailing space should be removed, admittedly it was like that before, but I only noticed it in this diff!

Tests/LinuxMain.swift
```diff
@@ -16,7 +16,22 @@
 
 import XCTest
 
-import SwiftCloudantTestSuite
+@testable import SwiftCloudantTests
 
-var tests: [XCTestCaseEntry] = [SwiftCloudantTestSuite.allTests()]
-XCTMain(tests)
+XCTMain( [
```
ricellis: This list of classes and the `allTests` var in each one are going to be a nuisance to curate, but I guess if that's the only option we have right now we just have to go with it.

rhyshort: There is a neat trick to have macOS builds check the test count, remind me sometime to dig it out.

Source/SwiftCloudant/Operations/Query/FindDocumentsOperation.swift
ianpartridge: Fixed.

Issue #170, opened 2018-02-12
Illegal Instruction error thrown when calling from command-line on OS X


jthomas:
I've found a bug which only surfaces when using the library from the command-line using `swift run` on OS X, rather than XCode.

Here's my `main.swift`...

```swift
import SwiftCloudant
import Foundation

let url = URL(string: "http://localhost:5984")!
let client = CouchDBClient(url:url, username:"username", password:"password")
let dbName = "todos"

let allDocs = GetAllDocsOperation(databaseName: "todos", rowHandler: { doc in
    print("Got document: \(doc)")
}) { response, info, error in
    if let error = error {
        print("error", error)
    } else {
        print("success")
    }
}

client.add(operation: allDocs)
CFRunLoopRun()
```

Here's the package manifest
```swift
// swift-tools-version:4.0
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "cloudant",
    dependencies: [
        // Dependencies declare other packages that this package depends on.
        .package(url: "https://github.com/cloudant/swift-cloudant.git", .revision("8bf984ef9f4040dcb50169c57769888f71f4fc38")),        
    ],
    targets: [
        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
        .target(
            name: "cloudant",
            dependencies: ["SwiftCloudant"]),
    ]
)
```

If I run the code using `swift run` the following error is thrown.

```
$ swift run
Compile Swift Module 'SwiftCloudant' (25 sources)
Compile Swift Module 'cloudant' (1 sources)
Linking ./.build/x86_64-apple-macosx10.10/debug/cloudant
Fatal error: Unexpectedly found nil while unwrapping an Optional value
Illegal instruction: 4
```

If I set up an xcode project for this test code and start the binary through that it works, which was super confusing! 

Digging into the library, I found the cause of the issue.
https://github.com/cloudant/swift-cloudant/blob/master/Source/SwiftCloudant/HTTP/URLSession.swift#L467-L468

```swift
bundleDisplayName = frameworkBundle.object(forInfoDictionaryKey: "CFBundleName") as! String
bundleVersionString = frameworkBundle.object(forInfoDictionaryKey: "CFBundleShortVersionString") as! String
```

`frameworkBundle.object` returns nil when running from the command-line.

I've opened a PR to resolve this issue.

[...]

PR #173, opened 2018-03-08
Use CouchDBClient.version for user-agent string

tomblench:
## What

Fix incorrect behaviour when fetching version number for user-agent string

## How

`Bundle(for:)` is not supported on Linux
(https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/Bundle.swift), and we don't expect it to be supported any time soon.

Instead, define the version number in CouchDBClient and hardcode the first part of the
user-agent as "SwiftCloudant".

## Testing

All existing tests pass

## TODO

Once merged, fix instructions in https://github.com/cloudant/cloudant-sync/blob/master/releasing-swift.md regarding version number updates.

## Issues

See #170

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,3 +1,5 @@
+// swift-tools-version:3.2
+
 //  Copyright (c) 2016 IBM Corp.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
@@ -8,6 +10,8 @@
 //  either express or implied. See the License for the specific language governing permissions
 //  and limitations under the License.
 //
+
+
 import PackageDescription
 
 let package = Package(
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,3 @@
-// swift-tools-version:3.1
 //  Copyright (c) 2016 IBM Corp.
 //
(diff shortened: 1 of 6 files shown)


Source/SwiftCloudant/CouchClient.swift
```diff
@@ -72,6 +72,9 @@ public class CouchDBClient {
     internal let password: String?
     internal let rootURL: URL
 
+    // The version number of swift-cloudant, as a string
+    static let version = "0.7.0"
```
ricellis: we should make this a `0.7.1-SNAPSHOT` in master and update at release time.

Package.swift
```diff
@@ -1,4 +1,3 @@
-// swift-tools-version:3.1
```
ricellis: I think we should leave this in, as this file is in the 3.x format

tomblench: Ah, so did we want `swift-tools-version:3.2` which it originally was?

ricellis: yeah that probably makes sense

ricellis: oh, ha, the version on Travis only has 3.1 I guess that's why I dropped the level originally.

PR #174, opened 2018-03-12
Update pod spec

sandmman:
- [x] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://github.com/cloudant/swift-cloudant/blob/master/DCO1.1.txt)
- [x] You have added tests for any code changes
- [x] You have updated the [CHANGELOG.md](https://github.com/cloudant/swift-cloudant/blob/master/CHANGELOG.md)
- [x] You have completed the PR template below:

## What

When using SwiftCloudant in a Swift 4.0 project, cocoapods would automatically set the swift version for the pod to 4.0. This would fail to build due to a non-exhaustive swift block. 

## How

I added the extra case statement to the switch block to remove the error/warning and set the swift version in the podspec to swift 3.2 to ensure it is compiled as expected

## Testing

I installed this cocoapod into another project and swift version was set as expected.


ricellis:
Thanks for this, we'll take a look at it as soon as we can.

